PR #38, opened 2017-06-09
Refactored health status provider

rodrigogs:
I just created separated functions for each health information, it makes code much cleaner and maintainable. Also, you dont need to feed an object through the promise chain.

Added editorconfig, since tab indentation is used in the project, but spaces is the default style in many editors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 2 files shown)


rodrigogs:
@icebob did it :)

rodrigogs:
Sorry, forgot one :P

PR #63, opened 2017-08-07
ServiceProxy helper

qdoop:
ServiceProxy could be many more things. This code basically is syntactic sugar.
EXAMPLE
let proxy_math=ServiceProxy(broker, "math")
proxy_math.add({a:1, b:2})


icebob:
Good idea, but please add relevant tests.

PR #72, opened 2017-08-13
AMQP Transporter

### LINKED ISSUE #1: AMQP transporter
Add AMQP support with [amqp.node](https://github.com/squaremo/amqp.node)

Nathan-Schwartz:
Resolves #1 

This Transporter is a bit more involved than some of the others because it needs to treat some packet types specially. Broadcasted packets need to use AMQP exchanges, but targeted packets only need a queue.  This implementation also deals with Requests a bit differently than other Transporters do. Each action has its own work queue that multiple nodes can pull messages from. This means that messages are less likely to be lost (due to crashes / etc). Each node will consume more messages when it has an availability. The number of messages that can be handled concurrently by a node can be specified with the `amqp.prefetch` option.

Requests and Responses don’t have a set time-to-live, but all other packet types do. Currently events’ time-to-live can be configured using the `amqp.eventTimeToLive` option, but the other types are hardcoded at 5 seconds.

**Transporter Options:**
```js
new AmqpTransporter({
  amqp: {
    url: "amqp://guest:guest@localhost:5672",
    eventTimeToLive: 5000,
    prefetch: 1   
  }
});
```

### AMQP Transport behaviour by topic

AMQP Transporter diagram

I’ve included unit tests and some integration tests. The integration tests are skipped right now because they require an amqp server to be running at `amqp://guest:guest@localhost:5672`. Let me know if I should remove them. :+1: 

Other notes:
- I don’t have a ton of experience with JSDoc, don’t hesitate to let me know if I can improve them. 
- I tried to to use `istanbul ignore` the same way as in other files, but let me know if I should change anything.

### DIFF
--- a/src/transit.js
+++ b/src/transit.js
@@ -175,30 +175,33 @@ class Transit {
 	 * @memberOf Transit
 	 */
 	makeSubscriptions() {
+		this.subscribing = Promise.all([
+			// Subscribe to broadcast events
+			this.subscribe(P.PACKET_EVENT),
 
-		// Subscribe to broadcast events
-		this.subscribe(P.PACKET_EVENT);
+			// Subscribe to requests
+			this.subscribe(P.PACKET_REQUEST, this.nodeID),
 
-		// Subscribe to requests
-		this.subscribe(P.PACKET_REQUEST, this.nodeID);
+			// Subscribe to node responses of requests
+			this.subscribe(P.PACKET_RESPONSE, this.nodeID),
 
-		// Subscribe to node responses of requests
-		this.subscribe(P.PACKET_RESPONSE, this.nodeID);
+			// Discover handler
+			this.subscribe(P.PACKET_DISCOVER),
 
-		// Discover handler
-		this.subscribe(P.PACKET_DISCOVER);
+			// NodeInfo handler
+			this.subscribe(P.PACKET_INFO), // Broadcasted INFO. If a new node connected
+			this.subscribe(P.PACKET_INFO, this.nodeID), // Response INFO to DISCOVER packet
 
-		// NodeInfo handler
-		this.subscribe(P.PACKET_INFO); // Broadcasted INFO. If a new node connected
-		this.subscribe(P.PACKET_INFO, this.nodeID); // Response INFO to DISCOVER packet
+			// Disconnect handler
+			this.subscribe(P.PACKET_DISCONNECT),
 
-		// Disconnect handler
-		this.subscribe(P.PACKET_DISCONNECT);
-
-		// Heart-beat handler
-		this.subscribe(P.PACKET_HEARTBEAT);
-
-		return Promise.resolve();
+			// Heart-beat handler
+			this.subscribe(P.PACKET_HEARTBEAT),
+		])
+			.then(() => {
+				this.subscribing = null;
+			});
+		return this.subscribing;
 	}
 
 	/**
@@ -498,6 +501,13 @@ class Transit {
 	 * @memberOf Transit
 	 */
 	publish(packet) {
+		if (this.subscribing) {
+			return this.subscribing
+				.then(() => {
+					this.stat.packets.sent = this.stat.packets.sent + 1;
+					return this.tx.publish(packet);
+				});
+		}
 		this.stat.packets.sent = this.stat.packets.sent + 1;
 		return this.tx.publish(packet);
 	}
--- a/src/transit.js
+++ b/src/transit.js
@@ -50,7 +50,7 @@ class Transit {
 		};
 
 		this.connected = false;
-		this.disconnecting = true;
+		this.disconnecting = false;
 
 		if (this.tx)
 			this.tx.init(this, this.messageHandler.bind(this), this.afterConnect.bind(this));
--- a/src/transporters/amqp.js
+++ b/src/transporters/amqp.js
@@ -79,12 +79,13 @@ class AmqpTransporter extends Transporter {
 					connection
 						.on("error", (err) => {
 							this.connected = false;
-							this.logger.error("AMQP connection error!", err);
+							reject(err);
+							this.logger.error("AMQP connection error!");
 						})
 						.on("close", (err) => {
 							this.connected = false;
-							const crashWorthy = require("amqplib/lib/connection").isFatalError(err);
-							this.logger.error("AMQP connection closed!", crashWorthy && err ||  "");
+							reject(err);
+							this.logger.error("AMQP connection closed!");
 						})
 						.on("blocked", (reason) => {
 							this.logger.warn("AMQP connection blocked!", reason);
@@ -106,11 +107,13 @@ class AmqpTransporter extends Transporter {
 							channel
 								.on("close", () => {
 									this.connected = false;
+									reject();
 									this.logger.warn("AMQP channel closed!");
 								})
-								.on("error", (error) => {
+								.on("error", (err) => {
 									this.connected = false;
-									this.logger.error("AMQP channel error!", error);
+									reject(err);
+									this.logger.error("AMQP channel error!");
 								})
 								.on("drain", () => {
 									this.logger.info("AMQP channel drained!");
@@ -121,13 +124,15 @@ class AmqpTransporter extends Transporter {
 						})
 						.catch((err) => {
 							/* istanbul ignore next*/
-							this.logger.error("AMQP failed to create channel!", err);
+							this.logger.error("AMQP failed to create channel!");
+							this.connected = false;
 							reject(err);
 						});
 				})
 				.catch((err) => {
 					/* istanbul ignore next*/
-					this.logger.error("AMQP failed to connect!", err);
+					this.logger.warn("AMQP failed to connect!");
+					this.connected = false;
 					reject(err);
 				});
 		});
--- a/src/transporters/amqp.js
+++ b/src/transporters/amqp.js
@@ -1,0 +1,329 @@
+/*
+ * moleculer
+ * Copyright (c) 2017 Ice Services (https://github.com/ice-services/moleculer)
+ * MIT Licensed
+ */
+
+"use strict";
+
+const Promise		= require("bluebird");
+const Transporter 	= require("./base");
+const {
+	PACKET_REQUEST,
+	PACKET_RESPONSE,
+	PACKET_UNKNOW,
+	PACKET_EVENT,
+	PACKET_DISCOVER,
+	PACKET_INFO,
+	PACKET_DISCONNECT,
+	PACKET_HEARTBEAT,
+} = require("../packets");
+
+/**
+ * Transporter for AMQP
+ *
+ * More info: https://www.amqp.org/
+ *
+ * @class AmqpTransporter
+ * @extends {Transporter}
+ */
+class AmqpTransporter extends Transporter {
+
+	/**
+	 * Creates an instance of AmqpTransporter.
+	 *
+	 * @param {any} opts
+	 *
+	 * @memberOf AmqpTransporter
+	 */
+	constructor(opts) {
+		if (typeof opts === "string")
+			opts = { amqp: { url: opts } };
+
+		// Number of requests a broker will handle concurrently
+		if (typeof opts.amqp.prefetch !== "number")
+			opts.amqp.prefetch = 1;
+
+		// Number of milliseconds before an event expires
+		if (typeof opts.amqp.eventTimeToLive !== "number")
+			opts.amqp.eventTimeToLive = 5000;
+
+		super(opts);
+
+		this.connection = null;
+		this.channel = null;
+		this.bindings = [];
+	}
+
+	/**
+	 * Connect to a AMQP server
+	 *
+	 * @memberOf AmqpTransporter
+	 */
+	connect() {
+		return new Promise((resolve, reject) => {
+			let amqp;
+			try {
+				amqp = require("amqplib");
+			} catch(err) {
+				/* istanbul ignore next */
+				this.broker.fatal("The 'amqplib' package is missing! Please install it with 'npm install amqplib --save' command!", err, true);
+			}
+
+			amqp.connect(this.opts.amqp.url)
+				.then(connection => {
+					this.connection = connection;
+					this.logger.info("AMQP connected!");
+
+					/* istanbul ignore next*/
+					connection
+						.on("error", (err) => {
+							this.connected = false;
+							this.logger.error("AMQP connection error!", err);
+						})
+						.on("close", (err) => {
+							this.connected = false;
+							const crashWorthy = require("amqplib/lib/connection").isFatalError(err);
+							this.logger.error("AMQP connection closed!", crashWorthy && err ||  "");
+						})
+						.on("blocked", (reason) => {
+							this.logger.warn("AMQP connection blocked!", reason);
+						})
+						.on("unblocked", () => {
+							this.logger.info("AMQP connection unblocked!");
+						});
+
+					connection
+						.createChannel()
+						.then((channel) => {
+							this.channel = channel;
+							this.onConnected().then(resolve);
+							this.logger.info("AMQP channel created!");
+
+							channel.prefetch(this.opts.amqp.prefetch);
+
+							/* istanbul ignore next*/
+							channel
+								.on("close", () => {
+									this.connected = false;
+									this.logger.warn("AMQP channel closed!");
+								})
+								.on("error", (error) => {
+									this.connected = false;
+									this.logger.error("AMQP channel error!", error);
+								})
+								.on("drain", () => {
+									this.logger.info("AMQP channel drained!");
+								})
+								.on("return", (msg) => {
+									this.logger.info("AMQP channel returned a message!", msg);
+								});
+						})
+						.catch((err) => {
+							/* istanbul ignore next*/
+							this.logger.error("AMQP failed to create channel!", err);
+							reject(err);
+						});
+				})
+				.catch((err) => {
+					/* istanbul ignore next*/
+					this.logger.error("AMQP failed to connect!", err);
+					reject(err);
+				});
+		});
+	}
+
+	/**
+	 * Disconnect from a AMQP server
+	 *
+	 * @memberOf AmqpTransporter
+	 * @description Close the connection and unbind this node's queues.
+	 * This prevents messages from being broadcasted to a dead node.
+	 * Note: Some methods of ending a node process don't allow disconnect to fire, meaning that
+	 * some dead nodes could still receive published packets.
+	 * Queues and Exchanges are not be deleted since they could contain important messages.
+	 */
+	disconnect() {
+		if (this.connection && this.channel && this.bindings) {
+			return Promise.all(this.bindings.map(binding => this.channel.unbindQueue(...binding)))
+				.then(() => this.channel.close())
+				.then(() => this.connection.close())
+				.then(() => {
+					this.bindings = null;
+					this.channel = null;
+					this.connection = null;
+				});
+		}
+	}
+
+	/**
+	 * Get assertQueue options by packet type.
+	 *
+	 * @param {String} packetType
+	 *
+	 * @memberOf AmqpTransporter
+	 */
+	_getQueueOptions(packetType) {
+		switch(packetType) {
+			// Requests and responses don't expire.
+			case PACKET_REQUEST:
+			case PACKET_RESPONSE:
+				return {};
+			// Packet types meant for internal use will expire after 5 seconds.
+			case PACKET_DISCOVER:
+			case PACKET_DISCONNECT:
+			case PACKET_UNKNOW:
+			case PACKET_INFO:
+			case PACKET_HEARTBEAT:
+				return { messageTtl: 5000 };
+			// Consumers can decide how long events live. Defaults to 5 seconds.
+			case PACKET_EVENT:
+				return { messageTtl: this.opts.amqp.eventTimeToLive };
+		}
+	}
+
+	/**
+	 * Build a function to handle requests.
+	 *
+	 * @param {String} cmd
+	 *
+	 * @memberOf AmqpTransporter
+	 */
+	_consumeCB(cmd) {
+		return (msg) => {
+			const result = this.messageHandler(cmd, msg.content);
+
+			// If a promise is returned, acknowledge the message after it has resolved.
+			// This means that if a worker dies after receiving a message but before responding, the
+			// message won't be lost and it can be retried.
+			if(cmd === PACKET_REQUEST) {
+				if (result instanceof Promise) {
+					return result.then(() => this.channel.ack(msg));
+				} else {
+					this.channel.ack(msg);
+				}
+			}
+		};
+	}
+
+
+	/**
+	 * Subscribe to a command
+	 *
+	 * @param {String} cmd
+	 * @param {String} nodeID
+	 *
+	 * @memberOf AmqpTransporter
+	 * @description Initialize queues and exchanges for all packet types except Request.
+	 *
+	 * All packets that should reach multiple nodes have a dedicated qeuue per node, and a single
+	 * exchange that routes each message to all queues. These packet types will not use
+	 * acknowledgements and have a set time-to-live. The time-to-live for EVENT packets can be
+	 * configured in options.
+	 * Examples: INFO (sometimes), DISCOVER, DISCONNECT, HEARTBEAT, EVENT
+	 *
+	 * Other Packets are headed towards a specific node or queue. These don't need exchanges and
+	 * packets of this type will not expire.
+	 * Examples: REQUEST, RESPONSE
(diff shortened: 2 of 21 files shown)


src/transporters/amqp.js
```diff
@@ -0,0 +1,329 @@
...
+	 * @param {any} opts
+	 *
+	 * @memberOf AmqpTransporter
+	 */
+	constructor(opts) {
+		if (typeof opts === "string")
+			opts = { amqp: { url: opts } };
+
+		// Number of requests a broker will handle concurrently
+		if (typeof opts.amqp.prefetch !== "number")
```
WoLfulus: This will break if opts is a string.

src/transporters/amqp.js
```diff
@@ -0,0 +1,329 @@
...
+			opts.amqp.prefetch = 1;
+
+		// Number of milliseconds before an event expires
+		if (typeof opts.amqp.eventTimeToLive !== "number")
```
WoLfulus: This too

src/transporters/amqp.js
Nathan-Schwartz: @WoLfulus If opts was a string it should have already been reassigned to an object by this point. The first if-statement in the constructor should handle this case, I think.

WoLfulus: I'm sorry, I didn't see you reassigning opts at line 41, so yeah, it wont fail.

My bad :(

Nathan-Schwartz: No worries :slightly_smiling_face:

icebob:
Thanks @Nathan-Schwartz . Great job! I'm on holiday now, but I'm going to review it next week.

src/transit.js
```diff
@@ -498,6 +501,13 @@ class Transit {
 	 * @memberOf Transit
 	 */
 	publish(packet) {
+		if (this.subscribing) {
```
icebob: Is it solve an issue that publish a packet before subscriptions are ready

src/transporters/amqp.js
```diff
@@ -0,0 +1,329 @@
...
+							this.connected = false;
+							this.logger.error("AMQP connection error!", err);
+						})
+						.on("close", (err) => {
+							this.connected = false;
+							const crashWorthy = require("amqplib/lib/connection").isFatalError(err);
+							this.logger.error("AMQP connection closed!", crashWorthy && err ||  "");
+						})
+						.on("blocked", (reason) => {
+							this.logger.warn("AMQP connection blocked!", reason);
```
icebob: What's mean bocker & unblocked?

src/transit.js
Nathan-Schwartz: Yes, thats correct.

test/integration/amqp/events/events.spec.js
```diff
@@ -0,0 +1,54 @@
...
+const findResponse = (arr) => arr
+	.map(string => {
+		const line = string
+			.split("\n")
+			.find(a => a.indexOf(commonString) > 0);
+		return line ? line.slice(0, line.indexOf(commonString)) : false;
+	})
+	.filter(a => a);
+
+xdescribe("Test AMQPTransporter events", () => {
```
icebob: What is `xdescribe`?

src/transporters/amqp.js
Nathan-Schwartz: From the docs I read that a RabbitMQ server (after version 3.2.0) can decide to block a connection. 

> Typically it will do this if there is some resource shortage, e.g., memory, and messages are published on the connection. - [amqp.node](http://www.squaremobius.net/amqp.node/channel_api.html#model_events)

I don't know how common this error is, but I thought logging would help with debugging.

test/integration/amqp/events/events.spec.js
Nathan-Schwartz: `xdescribe` is the same as `describe.skip`, I did this for the integration tests I wrote so the describe block won't run. These tests require an AMQP server to be running at `amqp://guest:guest@localhost:5672`, so I don't think they would pass in the current CI environment.

icebob:
Integration test is perfect. But in case of dependency we can't running it in `npm test`. Maybe we need to move out from jest scope and create an other npm script which can run it. But I'm going to consider it.
Tomorrow I'm going to test is with a RabbitMQ server.

By the way, very nice PR. Congratulation :wink: :+1: (I like that you followed my coding styles)

test/integration/amqp/events/events.spec.js
icebob: Thanks, I didn't know `xdescribe` :)

Nathan-Schwartz:
I think another script makes sense 👍 

Thank you! I really enjoyed working on Moleculer, it is an awesome project!

icebob:
@Nathan-Schwartz I'm trying to connect a remote RabbitMQ, but I got ECONNREFUSED from 127.0.0.1.
```js
// Create broker
let broker = new ServiceBroker({
	namespace: "multi",
	nodeID: process.argv[2] || "server-" + process.pid,
	transporter: "amqp://guest:guest@192.168.51.29:5672",
	logger: console
});
```

Err:
```
[2017-08-16T12:36:15.502Z] ERROR server-8120/TRANSPORTER: AMQP failed to connect! { Error: connect ECONNREFUSED 127.0.0.1:5672
    at Object.exports._errnoException (util.js:1018:11)
    at exports._exceptionWithHostPort (util.js:1041:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)
  cause:
   { Error: connect ECONNREFUSED 127.0.0.1:5672
       at Object.exports._errnoException (util.js:1018:11)
       at exports._exceptionWithHostPort (util.js:1041:20)
       at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)
     code: 'ECONNREFUSED',
     errno: 'ECONNREFUSED',
     syscall: 'connect',
     address: '127.0.0.1',
     port: 5672 },
  isOperational: true,
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5672 }
```

icebob:
I tried the `dev/amqp.js` with `let open = require("amqplib").connect("amqp://192.168.51.29");` and it's working, no connection error.

Nathan-Schwartz:
@icebob You are correct, I should have passed in `this.opts.amqp.url`, not `this.opts.amqp`. I am not quite sure how it was connecting successfully earlier 🤔. Anyways, I just made the change and added it to the first commit.

icebob:
Thanks. Btw, how you could modify an old commit? :)

icebob:
I tested. It seems it's working properly ;)

icebob:
A minor difference. In all other transporter if the connection lost, trying reconnect in 5 sec. But in amqp, 

NATS:
```
[2017-08-16T16:31:15.442Z] INFO  server-4064/BROKER: 'file' service is registered!
[2017-08-16T16:31:15.451Z] INFO  server-4064/BROKER: 'math' service is registered!
[2017-08-16T16:31:15.459Z] INFO  server-4064/TRANSIT: Connecting to the transporter...
[2017-08-16T16:31:15.509Z] INFO  server-4064/TRANSPORTER: NATS client is connected!
[2017-08-16T16:31:15.532Z] INFO  server-4064/BROKER: Broker started.
[2017-08-16T16:31:24.121Z] WARN  server-4064/TRANSPORTER: NATS client is disconnected!
[2017-08-16T16:31:26.122Z] WARN  server-4064/TRANSPORTER: NATS client is reconnecting...
[2017-08-16T16:31:29.129Z] WARN  server-4064/TRANSPORTER: NATS client is reconnecting...
[2017-08-16T16:31:32.155Z] WARN  server-4064/TRANSPORTER: NATS client is reconnecting...
[2017-08-16T16:31:35.158Z] WARN  server-4064/TRANSPORTER: NATS client is reconnecting...
[2017-08-16T16:31:38.181Z] WARN  server-4064/TRANSPORTER: NATS client is reconnecting...
[2017-08-16T16:31:38.184Z] INFO  server-4064/TRANSPORTER: NATS client is reconnected!
```

AMQP:
```
[2017-08-16T16:33:30.806Z] INFO  server-3328/BROKER: 'file' service is registered!
[2017-08-16T16:33:30.814Z] INFO  server-3328/BROKER: 'math' service is registered!
[2017-08-16T16:33:30.822Z] INFO  server-3328/TRANSIT: Connecting to the transporter...
[2017-08-16T16:33:30.946Z] INFO  server-3328/TRANSPORTER: AMQP connected!
[2017-08-16T16:33:30.956Z] INFO  server-3328/TRANSPORTER: AMQP channel created!
[2017-08-16T16:33:31.100Z] INFO  server-3328/BROKER: Broker started.
[2017-08-16T16:33:53.759Z] ERROR server-3328/TRANSPORTER: AMQP connection error! { Error: write ECONNRESET
    at exports._errnoException (util.js:1018:11)
    at Socket._writeGeneric (net.js:715:26)
    at Socket._write (net.js:734:8)
    at doWrite (_stream_writable.js:334:12)
    at writeOrBuffer (_stream_writable.js:320:5)
    at Socket.Writable.write (_stream_writable.js:247:11)
    at Socket.write (net.js:661:40)
    at Connection.C.sendMethod (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:531:17)
    at Object.accept (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:87:18)
    at Connection.mainAccept [as accept] (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:63:33)
    at Socket.go (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:476:48)
    at emitNone (events.js:86:13)
    at Socket.emit (events.js:185:7)
    at emitReadable_ (_stream_readable.js:432:10)
    at emitReadable (_stream_readable.js:426:7)
    at readableAddChunk (_stream_readable.js:187:13) code: 'ECONNRESET', errno: 'ECONNRESET', syscall: 'write' }
[2017-08-16T16:33:53.761Z] WARN  server-3328/TRANSPORTER: AMQP channel closed!
[2017-08-16T16:33:53.763Z] ERROR server-3328/TRANSPORTER: AMQP connection closed! { Error: write ECONNRESET
    at exports._errnoException (util.js:1018:11)
    at Socket._writeGeneric (net.js:715:26)
    at Socket._write (net.js:734:8)
    at doWrite (_stream_writable.js:334:12)
    at writeOrBuffer (_stream_writable.js:320:5)
    at Socket.Writable.write (_stream_writable.js:247:11)
    at Socket.write (net.js:661:40)
    at Connection.C.sendMethod (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:531:17)
    at Object.accept (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:87:18)
    at Connection.mainAccept [as accept] (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:63:33)
    at Socket.go (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\connection.js:476:48)
    at emitNone (events.js:86:13)
    at Socket.emit (events.js:185:7)
    at emitReadable_ (_stream_readable.js:432:10)
    at emitReadable (_stream_readable.js:426:7)
    at readableAddChunk (_stream_readable.js:187:13) code: 'ECONNRESET', errno: 'ECONNRESET', syscall: 'write' }
[2017-08-16T16:33:53.764Z] ERROR server-3328/TRANSPORTER: AMQP connection closed!

d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\channel.js:149
    throw new IllegalOperationError(msg, stack);
    ^
IllegalOperationError: Channel closed
    at Channel.<anonymous> (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\channel.js:149:11)
    at Channel.C.publish (d:\Work\JS\ice-services\moleculer\node_modules\amqplib\lib\channel_model.js:158:15)
    at AmqpTransporter.publish (d:\Work\JS\ice-services\moleculer\src\transporters\amqp.js:319:17)
    at Transit.publish (d:\Work\JS\ice-services\moleculer\src\transit.js:512:18)
    at Transit.emit (d:\Work\JS\ice-services\moleculer\src\transit.js:216:8)
    at ServiceBroker.emit (d:\Work\JS\ice-services\moleculer\src\service-broker.js:1122:17)
    at Timeout.setInterval [as _onTimeout] (d:\Work\JS\ice-services\moleculer\examples\multi-server\server.js:35:9)
    at ontimeout (timers.js:380:14)
    at tryOnTimeout (timers.js:244:5)
    at Timer.listOnTimeout (timers.js:214:5)
```
And the process exited.
Could you modify it to similar logic?

Nathan-Schwartz:
@icebob I do it using [`git commit --amend`](https://git-scm.com/docs/git-commit#git-commit---amend) if it I want to change the most recent commit. This will add staged changes to the last commit and let you change the commit message. Otherwise, I use [`git rebase -i <commit>`](https://git-scm.com/docs/git-rebase) to can squash, rename commits, and change old commits.

It can create problems if you change a commit that is already merged though :)

About reconnecting logic:

Nathan-Schwartz:
I wanted to implement reconnection, but its a bit tricky.

Unfortunately right now `amqp.node` doesn't handle connection recovery:
- https://github.com/squaremo/amqp.node/issues/153
- https://github.com/squaremo/amqp.node/issues/25

When the connection is lost, we could still try something like 
```js
setTimeout(() => this.connect(), 5000);
```
But the problem is that we don't know if all of the queues and exchanges were created successfully, and the exchanges might not be bound.

To make sure that all exchanges, queues, and bindings are ready we would need Transit to call `makeSubscriptions` again, and also send another `<prefix>.INFO` packet so we can call `_makeServiceSpecificSubscriptions` to initialize queues for requests.

If we assume that all queues, bindings, and exchanges are fine, reconnecting could work in some cases. I think there would be some edge-cases we couldn't handle well, though.

I tend to think that the transporter will behave more predictably if we don't try to reconnect.

What do you think?

icebob:
The reconnecting logic is implemented in transit, not in transporters. Just need to reject the connect promise correctly. I'm trying to fix it.

Nathan-Schwartz:
Ahh I think I understand now. Trying to reconnect when `connect` fails should be fine 👍 

I thought you were talking about reconnecting if the AMQP server goes down after `connect` succeeds.

Feel free to work on it, but I can fix it if you like.

icebob:
Could you check it? I blocked in it. After "AMQP channel closed!" I tried to call reject but it doesn't call catch in transit.
And I found a bug in constructor of transit `this.disconnecting = false;` It was true, which disable the reconnecting logic.

icebob:
Btw, this reconnecting solves both issue. Reconnecting if the first connect is failed, but it will also reconnect if connection lost after success connect. (in other transporters)

icebob:
Better, first connection fail is handler & reconnecting.

Nathan-Schwartz:
I think reconnecting after a failed initial connection is working for me.

I did notice one issue, though. It looks like if a message is sent before the transporter connects, the message is lost. Would it be possible to wait to publish messages until a connection is established? I think this could be done in transit.

My latest commit doesn't implement reconnecting if RabbitMQ goes down after the transporter connects successfully. For NATS and MQTT this is done by the client library, but amqp.node doesn't have this feature right now.

icebob:
Yes, I said wrong. Reconnecting is implemented in other libs because reject won't be called after it resolved.
So we skip it.

icebob:
So is it done? Can I merge?

Nathan-Schwartz:
I think it is done, but I want to make sure the implications for load balancing are okay 🙂 

Right now it will use prefetch to set the number of requests each node can handle concurrently. If prefetch is set to 2 a node will handle two messages at the same time from the queue. As soon as one request is completed, it will grab one more (and so on). The messages will wait in the queue until it is their turn and a node is free to handle the request.

I think this implementation could mean that Random and Round Robin strategies don't work as expected. The load balancing will still be done effectively, but I want to make sure this is okay before merging :+1:  

There is an integration test showing the expected behaviour [here](https://github.com/ice-services/moleculer/pull/72/files#diff-1da857f19c13c37aec52853d6e3f13d8R67)

icebob:
I read how the prefetch works. I think the default value 1 is working same as other transporter. But if somebody want to use other prefetch value, he/she can do it in transporter settings. So i think it's perfect solution.

Nathan-Schwartz:
Great! I think it is ready 🙂

icebob:
Merged. Very good job. Thanks again!

Further task: It would be good to separate amqp integration test, skip it in jest when we call `npm test` or `npm run ci`.  Dou you have any idea how we can do it?

Brainstorming: What do you think. Because of RabbitMQ has load balancing solution. This new AMQP transporter can work without Moleculer built-in load balancing? I think transporter disable load balancing in broker. But I have no experience with AMQP and I can't consider it completely :(

icebob:
@Nathan-Schwartz What do you think?

Nathan-Schwartz:
@icebob For the tests we could try a folder structure like this:
```
test
  |- unit
  `- integration
    |- internal
      |- broker-transit.spec.js
      |- serializer.spec.js
      `- (etc)
    |- thirdParty
      |- amqp
      `- (redis / nats / bull / etc)
```
and then we would ignore the thirdParty directory for the `ci` and `test` scripts. I'm not sure what the best solution is, I will keep thinking about it.

I think the AMQP transporter would work without Moleculer's load balancing. Are you thinking about making the transporter work with Moleculer load balancing, or disabling Moleculer's load balancing when the AMQP Transporter is used?

icebob:
Thanks @Nathan-Schwartz! 
1. I solved this issue. I renamed the *.spec.js files to *.spc.js, and added a new NPM script as "test:amqp" which is running only *.spc.js files. You can check it in `master`.

2. I think that some transporters can disable Moleculer's built-in load balancer if it has own load balancer like RabbitMQ or NATS Streaming.
It seems, I have to implement load balancing for EVENTS packets too to support Event Driven arch. And if I have to rewrite service registry maybe I will implement a switchable load balancer.

By the way within some days I will open a PR about AMQP optimization because it doesn't need to call `makeServiceSpecifigSubscriptions` in `publish` when send an INFO packet, because it can access to services if it call the `this.transit.getNodeInfo()`. This method returns a service list same as INFO packet contains. So it can also call it in `subscribe` method. Other benefits is the INFO packet will be able to send multiple times (in case of hot reload or dynamic service loading) and in this case it subscribes multiple times to the same queues.

Nathan-Schwartz:
@icebob Sounds good!

I didn't know about `this.transit.getNodeInfo`, that sounds like a great solution 🙂 

Currently it is re-asserting the REQ queues for every INFO packet, but if a service is loaded dynamically none of the new actions will a REQ queue yet. So I think `makeServiceSpecificSubscriptions` needs to be called every time a service is loaded.

One way to avoid re-asserting existing queues would be to track them in `this.queues` (similarly to `this.bindings`).

icebob:
The other problem if I destroy a service. How can I remove asserted REQ queues? Could you show an example?

Nathan-Schwartz:
@icebob I don't think the REQ queues or any exchanges should be destroyed. They could be in use by other services and destroying them would break the other nodes.

Note:
- If no Moleculer services were running with the AMQP transporter, the exchanges could be safely destroyed.
- Technically, all of the node-specific queues can be safely destroyed (for all packet types besides REQ).

If there was only one node for a service, a REQ queue could be destroyed without breaking things. 
If there were multiple nodes, you would need to wait until all of them went offline before deleting the queue. (Otherwise it would break the other nodes of this service.)

Even if it is possible to delete a queue without crashing nodes, I don't think we should do it. AMQP queues are often used as a Work Queue, so multiple nodes could be pulling from the same queue and there may be many messages waiting to be processed. If we delete the queue, we would delete all of the messages and this could be very problematic for a user.

icebob:
Thanks Nathan, It's good to know!

icebob:
I tried the `deleteQueue` in `disconnect` but I got "Channel ended" exceptions and I didn't know what was the reason. So now I set `autoDelete: true` to queues. It removes if no any consumers for the queue. It's working properly, but I don't know is it a good/best solution? @Nathan-Schwartz please review my changes in [amqp-impr](https://github.com/ice-services/moleculer/tree/amqp-impr) branch. 

https://github.com/ice-services/moleculer/compare/amqp-impr?expand=1#diff-b2879eb0ba697fd1aae416c88481d346

Thanks in advance!

Nathan-Schwartz:
@icebob It looks good 👍  

My only comment would be that I don't think we should use `autoDelete` for `PACKET_REQUEST` and `PACKET_RESPONSE`. I haven't tested this, but I think that it will delete the queue even if it still has messages.

It should be fine for the other packets since the messages will be removed after 5 seconds anyways.

icebob:
Ohh, I thought it would be deleted when it's empty and no consumers. :(
In this case I remove this attribute from REQ and RES queues. Thanks!

incubus8:
@Nathan-Schwartz please consider to use `Buffer.from` instead of `new Buffer`. It is being deprecated for security reason. 
as seen on NodeJS 6LTS doc - https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_new_buffer_array
cc @icebob

icebob:
Thanks. I changed it in [other branch](https://github.com/ice-services/moleculer/blob/amqp-impr/src/transporters/amqp.js#L313)

Nathan-Schwartz:
@icebob I just found out about the `ifEmpty` and `ifUnused` options for [`channel#deleteQueue`](http://www.squaremobius.net/amqp.node/channel_api.html#channel_deleteQueue). We could try to use this in disconnect for REQ and RES queues if you still want to.

icebob:
@Nathan-Schwartz It would be good. Could you create a PR?

icebob:
Hi @Nathan-Schwartz, I rewrote the AMQP integration tests. I used multiple broker & transporters in single files instead of multi processes. Here is my tests: https://github.com/ice-services/moleculer/tree/next/test/integration/amqp

What do you think, is it cover all test cases from your original tests?

Nathan-Schwartz:
@icebob Good idea! It looks good, but there are a few small things I'm noticing. Would you be willing to open a PR so it is easier to review / comment on?

icebob:
@Nathan-Schwartz Thanks. I already merged it, but I think you can add comments to the PR.

https://github.com/ice-services/moleculer/pull/106

Nathan-Schwartz:
@icebob Sorry it took me so long to do a proper review, I'll try to be faster in the future.

I started looking into the new AMQP Transporter and I think using the AMQP load-balancer and grouped events are cool ideas.

I wanted to ping you because I'm noticing some issues.

Two of the new tests for disabled load balancer fail:
* should retry unacked requests to other node
* Should use availability-based load balancing

The new integration tests don't include the following tests, and the old tests for these cases fail:
* Should not have a request or response dequeued until it has been received
* Should use availability-based load balancing
* Nodes should only receive one request at a time by default.

I tried to figure out why the tests fail and I found that we are no longer waiting for the action handler to complete before acknowledging messages. This means that `prefetch` is ineffective and AMQP can no longer be used to implement work queues. What was the reason for this change?

I am happy to help resolve the above issues and write more tests in the new style, if you are open to it.

Also, I want to suggest that we add a RabbitMQ server to Travis so that we can run integration tests before merging changes. Do you think that would be possible?

Thanks!
  Nathan

icebob:
Hi @Nathan-Schwartz,

thank you for your response. 
The main changes is that I removed promise returns from [messageHandler](https://github.com/ice-services/moleculer/blob/master/src/transit.js#L234). So the `ack` will be called after the message received to transit and not if the message processed & response sent. It seems it wasn't a good idea, but I can fix it. 

I would be happy if you can make the mentioned tests cases (would be fail). And after it I can start to fix the messageHandler code that the tests will be success.

Could you help me to write the test cases?

Nathan-Schwartz:
@icebob I see, thanks for the explanation.

Sounds like a plan! I will start working on the test cases today 👍 

What do you think about adding a RabbitMQ server to TravisCI?

icebob:
Great!

RabbitMQ on Travis would be good, but I want to skip amqp integration tests from `npm test`. So we could do it, if we change the script in travis.yml to `npm test && npm test:amqp`. Ok?

Nathan-Schwartz:
@icebob Sounds good to me!

Nathan-Schwartz:
@icebob Sometimes I am getting an error in the test case `Messages that haven't finished processing should be retryable by other nodes`. The error is `IllegalOperationError: Channel closing`. 

I looked into it for a while and it seems to happen when publishing the `DISCONNECT` packet from worker2 and worker3.

I tried turning off `autoDelete` for `DISCONNECT` packets and not using the `purge` function in case it was deleting the queue's before the test finished, but I wasn't able to make the test pass consistently.

It doesn't always fail but it could cause some inconvenience in CI, so I would be okay marking it as pending for now until we figure it out.

PR #75, opened 2017-08-14
WIP: Adds custom strategy to Service Registry

WoLfulus:
I wanted to dispatch calls to the node with the most recent heartbeat. But strategies are fixed and there's no way to set custom strategies.

This is still a WIP, so I'm opening it for discussions ;)

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,11 +7,8 @@
 "use strict";
 
 const {
-	STRATEGY_ROUND_ROBIN,
-	STRATEGY_RANDOM,
-
-	CIRCUIT_CLOSE, 
-	CIRCUIT_HALF_OPEN, 
+	CIRCUIT_CLOSE,
+	CIRCUIT_HALF_OPEN,
 	CIRCUIT_OPEN
 } = require("./src/constants");
 
@@ -29,10 +26,10 @@ module.exports = {
 
 	Errors: require("./src/errors"),
 
-	STRATEGY_ROUND_ROBIN,
-	STRATEGY_RANDOM,
+	RoundRobinStrategy: require("./src/strategies/roundrobin"),
+	RandomStrategy: require("./src/strategies/random"),
 
-	CIRCUIT_CLOSE, 
-	CIRCUIT_HALF_OPEN, 
+	CIRCUIT_CLOSE,
+	CIRCUIT_HALF_OPEN,
 	CIRCUIT_OPEN
 };
--- a/index.js
+++ b/index.js
@@ -26,8 +26,11 @@ module.exports = {
 
 	Errors: require("./src/errors"),
 
-	RoundRobinStrategy: require("./src/strategies/roundrobin"),
-	RandomStrategy: require("./src/strategies/random"),
+	Strategies: {
+		BaseStrategy: require("./src/strategies/base"),
+		RoundRobinStrategy: require("./src/strategies/roundrobin"),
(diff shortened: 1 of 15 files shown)


icebob:
Good solution. Now the strategy will be customizable :)

WoLfulus:
I'm thinking about removing the constants and making them a function. Also, I think I'll need to pass more info to the parameters. 

From the endpoint list how can I query the last heartbeat for example?

Depending on the strategy I'd like to have more information about the action.

icebob:
The registry doesn't store node info. You can access heartbeats via `broker.transit.nodes`

src/service-registry.js
```diff
@@ -420,24 +422,37 @@ class EndpointList {
...
-			// Round-robin
-
-			// Reset counter
-			if (this.counter >= this.list.length) {
-				this.counter = 0;
+		// Interface these strategies?
+		// RoundRobinStrategy, RandomStrategy, CustomStrategy(fn) ?
+		if (typeof this.opts.strategy === "function") {
+			const endpoint = this.opts.strategy(this.list);
+			if (!endpoint) {
```
WoLfulus: Oh, ok, I'll change this one ;)

test/unit/strategies/random.spec.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+
+let { RandomStrategy } = require("../../../src/strategies");
+
+expect.extend({
+	toBeAnyOf(received, expected) {
```
icebob: Good idea :)

icebob:
I like this new "classes" solution.

index.js
```diff
@@ -29,10 +26,10 @@ module.exports = {
 
 	Errors: require("./src/errors"),
 
-	STRATEGY_ROUND_ROBIN,
-	STRATEGY_RANDOM,
+	RoundRobinStrategy: require("./src/strategies/roundrobin"),
+	RandomStrategy: require("./src/strategies/random"),
```
icebob: Please put it under a "Strategies" prop like cachers, transporters, serializers with an index.js

src/strategies/roundrobin.js
```diff
@@ -0,0 +1,16 @@
+
+module.exports = class RoundRobinStrategy {
+
+	constructor() {
+		this.counter = 0;
+	}
+
```
icebob: Would be good an init base method (like in [other components](https://github.com/WoLfulus/moleculer/blob/a0d6f24e9dd3fe475f56564ba46421b5238b29fd/src/transporters/base.js#L37)), which will be called by ServiceRegistry. It will pass the `broker` to this function so you will have a pointer to `broker` or `nodes` list...etc

icebob: And maybe an abstract BaseStrategy class which implement this `init`, what you can override in your custom strategies :)

WoLfulus: Making those now!

index.js
WoLfulus: Done!

icebob: I don't see in the code.
I think
```js
Strategies: require("./src/strategies")
```
```

icebob: In index.d.ts is correct, but here is not. I think you forgot it :)

WoLfulus: Oh! Yeah, I missed it. Thanks!

Issue #87, opened 2017-08-23
`waitFor` method


icebob:
Create a `waitFor(serviceName: string|Array<string>): Promise` method in `Service` class.
With it you can wait for a service or multiple services to be available.

For example:

```js
module.exports = {
    name: "seed",
    
    started() {
        this.seedDB();
    },

    methods: {
        seedDB() {
            this.waitFor(["posts", "users"]).then(() => {
                //... Seed posts & users collections
            });
        }
    }
};
```

WoLfulus:
Awesome! :D

icebob:
It will be released in v0.11

Issue #86, opened 2017-08-22
Protocol v2 changes


icebob:
I'm going to update the protocol of Moleculer in the next version to support better events & other client solutions (in the future).
It will be a **breaking change**. The new version won't be able to communicate with the older versions.

## Common
- [x] add `ver` field to all packet, as version of protocol. Example: `ver: "2"` 

## `INFO`
- [x] remove `versions` field
- [x] remove `uptime` field
- [x] avoid double stringify in `services` & `events` fields if serializer is `JSON`
- [x] add `port` field (for tcp transporter in the future)
- [x] add `client` field
  ```js
  client: {
    type: "node", // Type of client "node" or "java" or "go"
    version: "0.11.0", // Version of client
    langVersion: "8.4.0" // version of Node or Java, or Go  
  }
  ```
- [x] add `config` field( for transfer settings in the future)
  ```js
  config: {
    compression: "gzip"
    encoding: "AES256"
    ...etc
  }
  ```

## `HEARTBEAT`
- [x] remove `uptime` field. _Reason: unneccessary info, you get it with `$node.health`_
- [x] add CPU usage _(to support cpu usage based strategy)_

## `REQUEST`
- [x] add `requestID` field from `ctx.requestID`
- [x] avoid double stringify in `params` & `meta` fields if serializer is `JSON`

## `RESPONSE`
- [x] avoid double stringify in `data` field if serializer is `JSON`
- [x] the `error` field is a JSON encoded string instead of a structure.

### `EVENT.<nodeID>` instead of `EVENT`
NodeID specified EVENT packet. It will be used for event groups & to reduce network traffic (send event to only nodes, which is subscribed to it). The old broadcasted `EVENT` packet is removed.
- [x] add `group` field for grouped events

### Ping (to measure the latency between nodes)
- [x] `PING` 
  ```js
  {
    time: 1503389454133 // GMT time of sender node
  }
  ```
- [x] `PONG` 
  ```js
  {
    time: 1503389454133, // GMT time of sender node
    arrived: 1503389454685 // GMT time of target node
  }
  ```
>The target node sends back its timestamp, so you can measure the time diffs between nodes.

PR #96, opened 2017-09-10
Balanced events & others

icebob:
## Breaking changes
- communication protocol changed #86
- `broker.bus` removed. Use `events` for subscriptions in service schema.
- `broker.emitLocal` renamed as `broker.broadcastLocal`
- `broker.emit` is balanced (grouping by service name)
- `broker.broadcast` is not balanced (same logic as the old `emit`)
- renamed internal events. (Starts with `$`)
  - `$node.connected`
  - `$node.updated`
  - `$node.disconnected`
  - `$services.changed`
  - `$circuit-breaker.closed`
  - `$circuit-breaker.opened`
  - `$circuit-breaker.half-opened`

- changed returned structure of `$node.list`, `$node.services`, `$node.actions`, `$node.health`
- `strategy` in broker options is a class instead of an instance

## Changes
- changed default broker options: `heartbeatInterval: 5`, `heartbeatTimeout: 15`, `circuitBreaker.maxFailures: 3`

## New
- new simple built-in logFormatter
- new service registry
- new internal action `$node.events`
- new `waitForServices` method in services & broker #87 
- ping other nodes `transit.sendPing`. For responses subscribe to `$node.pong` event.
- pluggable custom Validator
- new Error classes

## Performance
Local calls: [Performance: +4%]()


WoLfulus:
> strategy in broker options is a class instead of an instance

What if strategy is configurable?

Like if we want to pass parameters to the strategy to behave differently?

icebob:
@WoLfulus Sorry, but from now you can't pass instance to `strategy` because for the new registry needs a Strategy class to create many instances (every actions, every events).

If you want to create custom strategy with options I recommend to use the broker options.
```js
    registry: {
        strategy: MyCustomStrategy,
        strategyOptions: {
            a: 5
            // ...etc
        }
    }
})

class MyCustomStrategy {
    constructor(registry, broker) {
        super(registry, broker);
        this.options = broker.options.registry.strategyOptions; // Reach your custom settings
    }
}
```

PR #112, opened 2017-09-30
Support version on waitForServices

imatefx:
If a service has version ```waitForServices()``` function was not able to find the service because the version is undefined when passed to ```registry.hasService()```

it checks if the version can be extracted from the service names with format 
```
v1.accounts
v2.users
```

### DIFF
--- a/src/service-broker.js
+++ b/src/service-broker.js
@@ -667,7 +667,12 @@ class ServiceBroker {
 		return new Promise((resolve, reject) => {
 			const check = () => {
 				const count = serviceNames.filter(name => {
-					return this.registry.hasService(name);
+					// Split to version and name if possible
+					const svcMeta = name.split('.');
+					if(svcMeta.length === 2)
+						return this.registry.hasService(svcMeta[1], svcMeta[0].substr(1));
+					else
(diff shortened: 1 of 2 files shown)


icebob:
Thanks, but current solution works only with numeric version.
We should support all options, e.g.:
- `v3.users`
- `staging.users`
- `v2.order.saga`
...etc

Would be better to set an object if you want to define version.

**Normal current format**
```js
waitForServices("users");
waitForServices(["posts", "users"]);
```

**New optional versioned format**
```js
waitForServices({ name: "users", version: "staging" });
waitForServices([{ name: "order.saga", version: 2 }, {name: "users", version: "staging"]);
```

Both format should be supported.

Could you change it? If not, I will do it later.

imatefx:
ok, i will change it

imatefx:
I saw your edit only few minutes back 😄 , Added tests

icebob:
Nice job! Thank you! :+1:

imatefx:
😄  Thats my contribution to Open Source after a long time 👍

icebob:
Released in v0.11.2

PR #119, opened 2017-10-11
Update some default AMQP options

Nathan-Schwartz:
### Changes:
- Mark retry test as pending due to "random" failures.
- Allow all events to have timeout set with `eventTimeToLive`. Don't set a default `timeToLive`.
- Allow heartbeats to have a timeout specified. This is because it was possible for a heartbeat packet to expire before it was handled by the service, causing an erroneous "unavailable" state.
- Use a new connection for each purge. Closing the connection eliminates the need to use the `forceExit` with Jest
- Remove some unnecessary timeouts in the testing suite.
- Remove `messageTtl` option for internal packets, this could cause unexpected behaviours. `autoDelete` is a more elegant solution to the problem `messageTtl` was trying to solve in this case. (The queues and messages will be deleted when the node stops.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test": "jest --coverage --no-cache",
     "test:unit": "jest --testMatch \"**/unit/**spec.js\" --coverage --no-cache",
     "test:int": "jest --testMatch \"**/integration/**spec.js\" --coverage --no-cache",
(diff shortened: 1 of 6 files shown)


icebob:
Thank you!

Nathan-Schwartz:
@icebob Thanks for the quick release! 🙂

PR #126, opened 2017-10-21
Improve cacher

icebob:
## New cacher features
In action cache keys you can use meta keys with `#` prefix.
```js
broker.createService({
    name: "posts",
    actions: {
        list: {
            cache: {
                // Cache key:  "limit" & "offset" from ctx.params, "user.id" from ctx.meta
                keys: ["limit", "offset", "#user.id"],
                ttl: 5
            },
            handler(ctx) {...}
        }
    }
});
```

You can override the cacher default TTL setting in action definition.
```js
let broker = new ServiceBroker({
    cacher: {
        type: "memory",
        options: {
            ttl: 30 // 30 seconds
        }
    }
});

broker.createService({
    name: "posts",
    actions: {
        list: {
            cache: {
                // This cache entries will be expired after 5 seconds instead of 30.
                ttl: 5
            },
            handler(ctx) {...}
        }
    }
});
```

You can change the built-in cacher keygen function to your own one.
```js
let broker = new ServiceBroker({
    cacher: {
        type: "memory",
        options: {
            keygen(name, params, meta, keys) {
                // Generate a cache key
                return ...;
            }
        }
    }
});
```


TomKaltz:
@icebob I stumbled across https://github.com/adzerk/node-lru-native that might improve performance for this. Just wanted to pass it along.

icebob:
Thanks @TomKaltz

PR #129, opened 2017-10-26
more def improvements

rmccallum81:
More improvements but I have a comments below to help clarify

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ declare namespace Moleculer {
 		warn: (...args: any[]) => void;
 		info: (...args: any[]) => void;
 		debug?: (...args: any[]) => void;
-		trace: (...args: any[]) => void;
+		trace?: (...args: any[]) => void;
 	}
 
 	class LoggerInstance {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -5,7 +5,7 @@ declare namespace Moleculer {
 		warn: (...args: any[]) => void;
 		info: (...args: any[]) => void;
 		debug?: (...args: any[]) => void;
-		trace: (...args: any[]) => void;
+		trace?: (...args: any[]) => void;
```
rmccallum81: This is to make it compatible with `console` in node

index.d.ts
```diff
@@ -58,18 +66,19 @@ declare namespace Moleculer {
...
-	interface ServiceActionSchema {
-		cache?: boolean;
-		params?: object;
-		handler: (ctx: Context) => Promise<any>;
-	}
-
 	interface ServiceSettingSchema {
 		port?: number | string;
-		routes?: Route[];
+		routes?: RouteSchema[];
```
rmccallum81: Are port and routes not specific to the moleculer-web module? If so they should be in that project as an extension of this interface.

dani8art: Yes, It's true. I introduced this mistake, I thought `ServiceSettingSchema` has these two properties in `moleculer` by default. I'm sorry.

index.d.ts
```diff
@@ -58,18 +66,19 @@ declare namespace Moleculer {
...
+		transport?: string | Transporter;
```
icebob: `transport` is not a service setting as well.

index.d.ts
icebob: Please remove them.

index.d.ts
dani8art: Then, service setting has only: 
```
[name: string]: any;
```
is it true?

icebob: Yes, but there are two internal optional settings: https://moleculer.services/docs/service.html#Internal-settings

icebob:
Thanks!

PR #132, opened 2017-10-27
Expose BaseTransport and Transit classes

rmccallum81:


index.d.ts
```diff
@@ -471,6 +499,35 @@ declare namespace Moleculer {
 		}
 	}
 
+	class Transit {
+		constructor(broker: ServiceBroker, transporter: Transporter, opts: GenericObject);
```
rmccallum81: I wasn't able to determine what the `opts` parameter could contain, it can stay as a `GenericObject` or a better definition can be created

index.d.ts
```diff
@@ -471,6 +499,35 @@ declare namespace Moleculer {
...
+		afterConnect(wasReconnect: boolean): Promise<void>;
+		connect(): Promise<void>;
+		disconnect(): Promise<void>;
+		sendDisconnectPacket(): Promise<void>;
+		makeSubscriptions(): Promise<Array<void>>;
+		messageHandler(cmd: string, msg: GenericObject): boolean | Promise<void> | undefined;
```
rmccallum81: The docs say this should be just a `boolean` but I see more possibilities in the code, not sure if this is correct or should the code be corrected to only output boolean

index.d.ts
```diff
@@ -471,6 +499,35 @@ declare namespace Moleculer {
...
+		request(ctx: Context): Promise<void>;
+		sendEvent(nodeID: string, eventName: string, data: GenericObject): void;
+		sendBalancedEvent(eventName: string, data: GenericObject, nodeGroups: GenericObject): void;
+		sendEventToGroups(eventName: string, data: GenericObject, groups: Array<any>): void;
```
rmccallum81: I'm sure the `groups` param could have a better definition

index.d.ts
```diff
@@ -471,6 +499,36 @@ declare namespace Moleculer {
 		}
 	}
 
+	class Transit {
```
rmccallum81: I've made the assumption that all the function starting with '_' are internal only so I didn't include them.

index.d.ts
icebob: It's `Array<String>` or null/undefined

index.d.ts
```diff
@@ -471,6 +499,36 @@ declare namespace Moleculer {
...
+		sendPong(payload: GenericObject): Promise<void>;
+		processPong(payload: GenericObject): void;
+		sendHeartbeat(localNode: NodeHealthStatus): Promise<void>;
+		subscribe(topic: string, nodeID: string): Promise<void>;
+		publish(packet: Packet): Promise<void>;
+		serialize(obj: GenericObject, type: string): Buffer;
+		deserialize(buf: Buffer, type: string): any
+		deserialize(buf: Buffer): any
+	}
+
```
icebob: The Transit is an internal class, so you don't call them directly. So maybe it is not neccessary to publish in d.ts.

index.d.ts
icebob: Currently there is only one optional option `maxQueueSize: Number`

index.d.ts
icebob: It can be boolean or Promise

index.d.ts
rmccallum81: Thanks, I'll add that

index.d.ts
rmccallum81: I noticed that it can be used while creating a custom transport but perhaps an interface instead so it can't be instantiated.

icebob: It's true, transporter calls directly the transit.

icebob:
Can I merge?

rmccallum81:
Sure thing                                                                                                                                                                         Sent from my BlackBerry - the most secure mobile device                                                                                                                                                                           From: notifications@github.comSent: October 27, 2017 11:44 AMTo: moleculer@noreply.github.comReply-to: reply@reply.github.comCc: rmccallum81@gmail.com; author@noreply.github.comSubject: Re: [ice-services/moleculer] Expose BaseTransport and Transit classes (#132)  Can I merge?

icebob:
Thanks!

icebob:
@rmccallum81 Could you improve [MoleculerErrors](https://github.com/ice-services/moleculer/blob/master/index.d.ts#L468) in d.ts? Would be good to define different constructor arguments.

rmccallum81:
Sure can, I'll go over and submit something over the weekend.                                                                                                                                                                         Sent from my BlackBerry - the most secure mobile device                                                                                                                                                                           From: notifications@github.comSent: October 28, 2017 11:39 AMTo: moleculer@noreply.github.comReply-to: reply@reply.github.comCc: rmccallum81@gmail.com; mention@noreply.github.comSubject: Re: [ice-services/moleculer] Expose BaseTransport and Transit classes (#132)  @rmccallum81 Could you improve MoleculerErrors  in d.ts? Would be good to define different constructor arguments.

PR #134, opened 2017-10-30
Add call options type

rmccallum81:
Not sure if the `FallbackResponseHandler`  promise should be returning `FallbackResponse` type


icebob:
The `any` is correct, because it can also return Buffer or something...

PR #135, opened 2017-11-08
Improvement over actions params schema

dani8art:
Hi @icebob, @rmccallum81

I've refactored this definitions because the previous one force you to have only parameters with simple schema such as: `'string'`, `' number'` and `'object'`. In order to support more complex schema, for example: `{type:'number', min: 2, optional: true}`, and to support diferents *Validators* I've modified it to type `any`.

I my opinion, `ActionParamsSchema` doesn't depend on **moleculer**, it depend on the *Validator* you are using and should be typed in the module of this validator, because of that, **moleculer** should left it in blank with the type any.

What do you think?

BR.


icebob:
I think it is good. @rmccallum81 ?

rmccallum81:
The reason that I set it with those types is that the params are parsed from incoming data. As such it isn't feasible to set it to anything other than those listed with the exception of also adding "array". Keeping in mind also that the type here is actually `string` but limited to the options listed. 

If there is a possibility that the param could be a function or object, an extension of the `ActionParamTypes` could be made to allow for more flexibility. I'll add some notes shortly.

dani8art:
But, `ActionParamsSchema` must be a data `schema`, the incoming data will be checked and must compliance with this `schema`. For example if you have the following Action definition:
```
 name: 'get',
{
  params:{
    limit: {type: 'number', max: 100, optional: true}, 
    page: {type:'number', optional:true} 
  },
  handler(ctx) =>{
    return ...
  }
}
``` 
When the action is called: 
```
  ctx.call('get',{limit:10,page:1}); //well
  ctx.call('get',{limit:200,page:1}); //fail
  ctx.call('get'); //well
```
the object with parameters will be checked.

How do you do this with the current   `ActionParamsTypes` @rmccallum81 ? 
Is this how it should work @icebob?

icebob:
Yes, your example is correct.

rmccallum81:
If the params must use this format then the `ActionParamSchema` would look like this:
```
type ActionParamTypes = "number" | "string" | "object" | "array"; // Added array as it is a posibility
type ActionParam = {
    type: ActionParamTypes;
    optional?: boolean;
    [key: string]: any; // Any known params should also be defined before this
}
type ActionParamsSchema = { [key: string]: ActionParam };
```
If it is also possible to set a shortcut of the param to just the type then `ActionParamsSchema` becomes
`type ActionParamsSchema = { [key: string]: ActionParamTypes | ActionParam };`

dani8art:
Ok, I see your point @rmccallum81, but what happen if we decide to use other `Validator` that it uses other `schema`. How do you knon what `Validator` the developers will use? For example Joi:
```
{
  params: Joi.object().keys({
    name: Joi.string().min(4).max(30).required()
  }),
...
}
```
For this reason, I said beafore: "I my opinion, `ActionParamsSchema` doesn't depend on **moleculer**, it depend on the **Validator** you are using and it should be typed implementation of this `Validator`, because of that, **moleculer** should left it in blank with the type any."

https://moleculer.services/docs/validation.html

rmccallum81:
In this case the output from the `Joi.object().keys()` function matches the definition I provided above which will validate fine for the type

dani8art:
OK, let's imagine `Joi` always matches the definition. what's about others?: 

- ajv: https://www.npmjs.com/package/ajv

- jsonschema: https://www.npmjs.com/package/jsonschema

- mongoose-unique-validator: https://www.npmjs.com/package/mongoose-unique-validator

We shouldn't couple `moleculer` to any `Validator`. And the unique way to do that is that we use type `any`.

@icebob what do you think? what option would you want to use?

rmccallum81:
I see your point and agree that it should be able to use any validator, my attempt on the typing is to give developers (especially those new to moleculer) a way to easily know what format data should be in order to work with the framework. I would like to keep the definition available for the default validator params as it provides a quick and definite setup. The `ActionParamsSchema` can have a type of `type ActionParamsSchema = { [key: string]: any };` but keep the other ones to allow for easy development knowledge. Perhaps renaming to include Default or put it in the default validator definition as a sub-type

dani8art:
Ok, I agree with you. If I understand, you want to do something like:

```
type ActionParamsSchema = { [key: string]: any };
type ActionParamTypes = "boolean" |  "number" | "string" | "object" | "array" | ActionParamsSchema;
type ActionParams = {[key: string]: ActionParamTypes};
```

rmccallum81:
That can work.

icebob:
So the latest solution is good for everybody? Can I merge?

PR #151, opened 2017-12-18
update Code Climate ESLint to use eslint-4

efueger:


icebob:
Thank you. Now it's working again :+1:

PR #153, opened 2017-12-20
sub key nats is not being passed in.

disarticulate:
Fixes nats options passing. Unsure what other things are going on, but the _resolveTransporter method in the service broker is passing in ```opt.options``` and not ```opt.nats```


icebob:
It's working at present. Ref: https://github.com/ice-services/moleculer/issues/152#issuecomment-353009472

PR #166, opened 2018-01-09
ctx.meta into metrics event payload

dani8art:
Hi @icebob,

I've just added this commit to my fork of moleculer, and I think you could want to merge it into moleculer.

For my team is very important to know what there are into ctx.meta in a call, due to that, I've added it to the metrics event payload. 

BR.


icebob:
Thank you. Currently, I don't merge it because I'm thinking to extend the metrics logic to able to send data from `params` too.

dani8art:
Ok, I really like this, I had added the `params` too, but I thought you don't like to add `params` too. In my last commit, I've updated my changes. 

BR

icebob:
Thanks, great! But I would like to control it. 
For example, I have a service which has an import action, where I load data from a big JSON object, but I don't want to attach the whole big `params` to metrics. 
My suggestion is, we can enable/disable them in action definition.

```js
module.exports = {
    name: "test",
    actions: {
        import: {
            cache: true,
            metrics: {
                // Disable to add `ctx.params` to metrics payload.
                params: false,
                // Enable to add `ctx.meta` to metrics payload.
                meta: true
            },
            handler(ctx) {
                // ...
            }
        }
    }
}
```

Could you add this logic as well? You have a pointer to action definition via `ctx.action`.

Otherwise, I don't know what are the good default values. If both is `true` or `false`, or even `metrics: true`, but `params: false`? What do you think?

dani8art:
Ok, that sound so good.
As you said, I think metrics can be true and `params: false` by default. I'm thinking about it, and we can give also the possibility of pass a `function` in order to map your data. For example:

```javascript
metrics: {

   params: (obj)=>{
       return {obj.prop1, obj.prop2};
   },

   // Enable to add `ctx.meta` to metrics payload.
   meta: true
}
```
or maybe a list of field that you want to send on metrics object:

```javascript
metrics: {

   params: [ 'prop1', 'prop2' ],

   // Enable to add `ctx.meta` to metrics payload.
   meta: true
}
```

I think the best way is to have the three options: `boolean`, `function` and `array` in the both: meta and params.

What do you think?

icebob:
@dani8art Awesome idea! I think it is completely cover all demands :)

Could you implement it as well?

dani8art:
Yes, I’ll do it! ;)

dani8art:
Hi @icebob,

I've just finished the functionality of extra metrics, I've tried to do all according to your style line and also I've implemented tests for this new functionality. 

I would like to sum up my changes and explain them.

1. I've created two methods in the class Context.
```
_processExtraMetrics(payload);
```
this method calls the assign method with all extra metrics.
```
_assignExtraMetrics(name, payload)
``` 
this other method is called by the first one in order to decide between the three options of configuration: `boolean`, `array` or `function`.

2. I've implemented some tests for this two new method.

3. I've added the options by default in Service._createAction for `metrics` and I've modified the tests of this class.

4. I've implemented TypeScript support for this new functionality.

With this changes, you can do:

```
{
  name: "hello",
  params: {
    name: "string",
    lastname: "string"
  }
  handler(ctx)=>{
    return `Hello ${ctx.params.name} ${ctx.params.lastname}`;
  }
}
```
**Return all fields**

```
# action def.
{
  ...
  metrics: { params: true }
  ...
}

# metrics payload
{
  ...
  params: { name: "Jhon", lastname: "Rick" }
  ...
}
```

**Return only selected fields**

```
# action def.
{
  ...
  metrics: { params: [ "name" ] }
  ...
}

# metrics payload
{
  ...
  params: { name: "Jhon" }
  ...
}
```

**Return mapped fields**

```
# action def.
{
  ...
  metrics: { params: (p)=>p.name + ' ' + p.lastname }
  ...
}

# metrics payload
{
  ...
  params: "Jhon Rick"
  ...
}
```
Please review my code and you don't hesitate to ask me modifications or let me know if you have any doubt.

BR.

icebob:
Great! I'm reviewing...

src/context.js
```diff
@@ -299,6 +307,48 @@ class Context {
...
+		let toAssign = {};
+		if (_.isBoolean(def) && def) {
+			toAssign[name] = this[name];
+			_.assign(payload, toAssign);
+		} else if (_.isArray(def)) {
+			toAssign[name] = _.pick(this[name], def);
+			_.assign(payload, toAssign);
+		} else if (_.isFunction(def)) {
+			toAssign[name] = def(this[name]);
+			_.assign(payload, toAssign);
```
icebob: The context metrics is a critical part because it is called for every request/call. So it should be high performance.
I measured and the performance is decreased by -33%.
Master:
```
Suite: Call with statistics & metrics
√ No statistics*                    1,388,993 rps
√ With metrics*                       523,056 rps
√ With metrics & statistics*          475,057 rps
```

This PR:
```
Suite: Call with statistics & metrics
√ No statistics*                    1,492,456 rps
√ With metrics*                       359,542 rps
√ With metrics & statistics*          337,990 rps
```

I think due to `_.assign`. The `lodash` functions are not too good for performance. Could you change it to `Object.assign`?

src/service.js
```diff
@@ -245,6 +246,9 @@ class Service {
 		action.cache = action.cache !== undefined ? action.cache : (this.settings.$cache || false);
 		action.handler = this.Promise.method(handler.bind(this));
 
+		// action metrics options by default
+		action.metrics = defaultsDeep({ params: false, meta: true });
```
icebob: `defaultsDeep` is necessary? I think it can work without it as well.

dani8art: Sorry, this is a mistake, it would be as follow:

defaultsDeep(action.metrics, {params: false,meta: true});

In order to guarantee the defaults value are these values for all actions.

src/context.js
dani8art: Ok I’ll change it

icebob:
@dani8art do you need help from me?

dani8art:
hi, @icebob, sorry for my delay but I have not enough time last days.

I've done al changes you've requested me. Moreover, I have good news for you. I've reviewed my code in order to optimize and `_.assign` of `lodash` is not necessary. 

I hope all is well.

icebob:
No problem, thank you for your work. I will check and merge if everything is good.

icebob:
Thank you @dani8art for your contribution. I think it is a useful feature for metrics & monitoring!

Issue #167, opened 2018-01-09
Built-in clustering option in Moleculer Runner


icebob:
```sh
$ moleculer-runner --hot --instances 4
```

Same solution: https://github.com/ice-services/moleculer/blob/master/examples/loadtest/clients.js

@tinchoz49

tinchoz49:
I will try to do a PR today :)

tinchoz49:
@icebob today we talk about a way to run workers with different nodeID. Well, in my project I have a lot of moleculer nodes in different docker containers and I generate a nodeID for each node using node-uuid.

What I think is that maybe we can define a way to generate a nodeID when we use `--instances` flag.

For example, in one of the docker container runs `moleculer-runner --instances 4 services` and it gives you a nodeID:

df7cca36-3d7a-40f4-8f06-ae03cc22f045-1
df7cca36-3d7a-40f4-8f06-ae03cc22f045-2
df7cca36-3d7a-40f4-8f06-ae03cc22f045-3
df7cca36-3d7a-40f4-8f06-ae03cc22f045-4

where the nodeID is a `uuidv4` that represents something like the NODE_UNIQUE_ID + `worker.id` instance.

icebob:
I don't think it is a good approach because I want to keep the opportunity that the developer can define it.
My idea is that we use the nodeID from `config` and add the `worker.id` as a suffix like you did `xxxxx-1`. 

### Cases:
**User defined nodeID:**
```js
module.exports = {
  nodeID: "my-node"
}
```
Instance IDs:
```
my-node-1
my-node-2
my-node-3
my-node-4
```

**Your uuid approach**:
```js
module.exports = {
  nodeID: uuid.v4()
}
```
Instance IDs:
```
df7cca36-3d7a-40f4-8f06-ae03cc22f045-1
df7cca36-3d7a-40f4-8f06-ae03cc22f045-2
df7cca36-3d7a-40f4-8f06-ae03cc22f045-3
df7cca36-3d7a-40f4-8f06-ae03cc22f045-4
```

**Not defined ID**
In this case we use the [default broker logic](https://github.com/ice-services/moleculer/blob/master/src/utils.js#L35), get the computer name as nodeID.
```js
module.exports = {
  nodeID: null
}
```
Instance IDs:
```
my-computer-1
my-computer-2
my-computer-3
my-computer-4
```

sjmcdowall:
[...]

PR #169, opened 2018-01-11
Added built-in clustering option in Moleculer Runner #167

tinchoz49:
The goal of this PR is to add a new flag for Moleculer Runner, `--instances or -i` to offer support for clustering in Moleculer.

Some observations:

1. The default value for `--instances` is the count of CPUs available.
2. In production `(process.env.NODE_ENV === 'production')` if some worker is disconnected, moleculer-runner restart it automatically.
3. If the developer use the `--repl` flag only the worker n1 is going to run the repl.
4. During the start each worker node concat the `worker.id` to the nodeID defined.

### DIFF
--- a/bin/moleculer-runner.js
+++ b/bin/moleculer-runner.js
@@ -8,6 +8,7 @@
 "use strict";
 
 const Moleculer 	= require("../");
+const utils			= require("../src/utils");
 const fs 			= require("fs");
 const path 			= require("path");
 const _ 			= require("lodash");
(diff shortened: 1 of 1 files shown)


icebob:
Nice job, thank you!

However, I made a fix because it throws error if you run it without `--instances` because in this case, the `worker` is `undefined` :)

tinchoz49:
wow :sweat_smile: I didn't notice until now when I try run moleculer without instances, thanks for fixed it!

PR #193, opened 2018-02-06
Add generic definition of certain calls

rmccallum81:
This allows the developer to assign an output type for the calls
i.e. `broker.call<number>("test.add", { x: 1, y: 2 })` will return `Bluebird<number>` instead of `Bluebird<any>`


icebob:
@dani8art, please review.

dani8art:
Hi, I think it is correct.
Regards!

Issue #191, opened 2018-02-01
Joi validator error when use a Transport


ryam4u:
I get the the following error when i use Joi Validator and a Transport (NATS)

`Unhandled rejection TypeError: Converting circular structure to JSON`

icebob:
Thanks, It means Moleculer can't work with Joi because Joi object is not serializable.

PR #212, opened 2018-03-04
Prevent TypeError when serializing objects with circular references

rmccallum81:
Prevents issues like #191 by replacing circular references with a string when serializing requests and responses.

### DIFF
--- a/src/serializers/json.js
+++ b/src/serializers/json.js
@@ -33,7 +33,16 @@ class JSONSerializer extends BaseSerializer {
 	 * @memberof Serializer
 	 */
 	serialize(obj) {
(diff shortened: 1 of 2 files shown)


icebob:
I have been run a benchmark test on your code. I know it protects the circular objects but it has bad performance (-60%). 
image
It is a significant difference. So I don't merge it. Perhaps create a new JSON serializer class.

rmccallum81:
Not worth the performance hit

icebob:
I extracted it to a Gist, if somebody needs it.

https://gist.github.com/icebob/fe0c35195735f0ac6438a5d30979197a

Issue #218, opened 2018-03-10
Support Stream responses


icebob:
Would be good that Moleculer supports streams in action responses. It means you can return a `Stream` object in actions and Broker transfers the stream chunks while the stream is not closed. You can use it to send video or audio streams back to the caller or to API gateway.

**Required changes:**

- [ ] **change protocol**
  - [ ] add `stream: boolean` field to RESPONSE packet.
- [ ] **change `_responseHandler` in transit to handle streams**
  - [ ] check the `stream` field in the response. If `true`, (and not exists yet) create a WritableStream and return it to the caller.
  - [ ] if `false` and has opened stream, close it,
- [ ] **change `sendResponse` in transit to handle streams**
  - [ ] if `data`is a `ReadableStream` set `stream: true` in the packet payload
  - [ ] subscribe to `data` event and if fired, send a new response to the caller with the same contextID
  - [ ] subscribe to `end` event and if fired, send a new response with `stream: false` response with the same contextID
  - [ ] subscribe to `error` event and if fired, send a new response with the error to the caller with the same contextID

molobala:
That will be realy interresting ! Since actually I don't find no way to exchange stream between two services .

baadc0de:
Would this also allow to send stream as parameter? That would be very cool!

icebob:
Unfortunately not. It is just for responses. I'm thinking on it how I can add it to request side too. It is more complicated because on request side the node is selected dynamically (in addition in case of `disableBalancer` there is no node selection). But in response side, the target node is known.

icebob:
## UPDATE!

It will support request streams too. So you can send stream as `ctx.params` and you can return a stream.

**Example**
Here is a sample code which is working. It creates an `aes` service which has `encrypt` and `decrypt` action. 
With the `client` broker I open a file, call `aes.encrypt`, the response (which is a stream too) is sent to `aes.decrypt`. And this response (which is a stream again) is saved to a file and checked the result with SHA.
So it sends streams on both sides, multiple times.

```js
"use strict";

const ServiceBroker = require("../src/service-broker");
const fs = require("fs");
const path = require("path");
const crypto = require("crypto");

const password = "moleculer";

// Create broker #1
const broker1 = new ServiceBroker({
    nodeID: "client-" + process.pid,
    transporter: "NATS",
    logger: console,
    logLevel: "info"
});


// Create broker #2
const broker2 = new ServiceBroker({
    nodeID: "encrypter-" + process.pid,
[...]

PR #224, opened 2018-03-20
Add stream support in actions response

molobala:
As demanded and since I needed it , I tried to add stream support in action response, I just followed what you explained @icebob in #218   
Please take a look at and get me back

### DIFF
--- a/src/transit.js
+++ b/src/transit.js
@@ -352,7 +352,7 @@ class Transit {
 			let writable = this.pendingStreams.get(id);
 			if(!packet.stream){
 				//end of  stream
-				if(writable) writable.close();
+				if(writable) writable.end();
 				// Remove pending request
 				this.removePendingRequest(id);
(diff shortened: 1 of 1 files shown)


icebob:
Nice PR. I will check it, but I need more time now.
Otherwise, I don't want to merge it to the master yet. So I will create a new branch for this.

icebob:
Please close this PR and open under the [streams](https://github.com/moleculerjs/moleculer/tree/streams) branch instead of master. Thanks.

molobala:
Good  

I think it will be better too if we could stream in context object  
actually I have some actions that need to work with uploaded files, I did it , but it's not optimal, cause I save file first in local in Gateway API node, then , my Gateway expose public path to the file and send this one to my action that then download the file (much process) 

Thank you!

icebob:
I thinking same things, that we need streams in requests too, not only responses.

PR #225, opened 2018-03-20
Add stream support in actions response

molobala:
As demanded and since I needed it , I tried to add stream support in action response, I just followed what you explained @icebob in #218
Please take a look at and get me back

### DIFF
--- a/src/transit.js
+++ b/src/transit.js
@@ -352,7 +352,7 @@ class Transit {
 			let writable = this.pendingStreams.get(id);
 			if(!packet.stream){
 				//end of  stream
-				if(writable) writable.close();
+				if(writable) writable.end();
 				// Remove pending request
 				this.removePendingRequest(id);
 				this.pendingStreams.delete(id);
@@ -578,9 +578,10 @@ class Transit {
 		if(data && typeof data.on === 'function' && typeof data.read === 'function' && typeof data.pipe === 'function'){
 			//readable
 			payload.stream = true;
-			payload.data
+			data
 			.on('data',(chunck)=>{
 				payload.stream = true;
+				payload.data = chunck;
 				return this.publish(new Packet(P.PACKET_RESPONSE, nodeID, payload))
 					.catch(err => this.logger.error(`Unable to send '${id}' response to '${nodeID}' node.`, err));
 			})
@@ -590,6 +591,7 @@ class Transit {
 					.catch(err => this.logger.error(`Unable to send '${id}' response to '${nodeID}' node.`, err));
 			})
 			.on('error',(e)=>{
+				payload.stream = false;
 				payload.error=e;
 				return this.publish(new Packet(P.PACKET_RESPONSE, nodeID, payload))
 					.catch(err => this.logger.error(`Unable to send '${id}' response to '${nodeID}' node.`, err));
--- a/src/transit.js
+++ b/src/transit.js
@@ -577,20 +577,20 @@ class Transit {
 		}
 		if(data && typeof data.on === 'function' && typeof data.read === 'function' && typeof data.pipe === 'function'){
 			//readable
-			playload.stream = true;
+			payload.stream = true;
 			payload.data
 			.on('data',(chunck)=>{
-				playload.stream = true;
+				payload.stream = true;
 				return this.publish(new Packet(P.PACKET_RESPONSE, nodeID, payload))
 					.catch(err => this.logger.error(`Unable to send '${id}' response to '${nodeID}' node.`, err));
 			})
 			.on('end',()=>{
-				playload.stream = false;
+				payload.stream = false;
 				return this.publish(new Packet(P.PACKET_RESPONSE, nodeID, payload))
 					.catch(err => this.logger.error(`Unable to send '${id}' response to '${nodeID}' node.`, err));
 			})
 			.on('error',(e)=>{
-				playload.error=e;
+				payload.error=e;
 				return this.publish(new Packet(P.PACKET_RESPONSE, nodeID, payload))
 					.catch(err => this.logger.error(`Unable to send '${id}' response to '${nodeID}' node.`, err));
 			})
--- a/src/transit.js
+++ b/src/transit.js
@@ -13,6 +13,8 @@ const P				= require("./packets");
 const { Packet }	= require("./packets");
 const E 			= require("./errors");
 
+const {Writable, Readable} = require('stream');
+
 /**
  * Transit class
  *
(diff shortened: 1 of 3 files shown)


src/transit.js
```diff
@@ -337,17 +339,44 @@ class Transit {
...
 
 		// Update nodeID in context (if it uses external balancer)
 		req.ctx.nodeID = packet.sender;
 
 		// Merge response meta with original meta
 		_.assign(req.ctx.meta, packet.meta);
-
+		//Stream case
+		//get the underlined stream for id
+		if(packet.stream !== undefined && packet.stream !== "undefined"){
```
icebob: Is it necessary? `packet.stream` is defined as boolean. So it can be `true`, `false` or `undefined` if the other side is not supported it. So I think it can be just `if (packet.stream && packet.data)`

src/transit.js
```diff
@@ -337,17 +339,44 @@ class Transit {
...
+				this.pendingStreams.delete(id);
+			}
+			if(packet.stream && pass){
+				//on stream chunk
+				pass.write(packet.data.type === "Buffer" ? new Buffer.from(packet.data.data):packet.data);
+			}
+			else{
+				//create a new pass stream
+				pass = new Transform({
+					transform:function(chunck,encoding,done){
```
icebob: Please fix typo from `chunck` to `chunk`

src/transit.js
molobala: In my thought if `stream === undefined` that means the response has no stream , if we do just `packet.stream && packet.data` when to close the stream since `false` and `undefined` will be treated as the same ! 

I think we could just do `if(packet.stream !== undefined){` since stream is boolean like you said.

src/transit.js
molobala: sure !

src/transit.js
icebob: Ok, I see.  `if (packet.stream !== undefined) {` will be good.

icebob:
I see still `chunck` :) The correct is `chunk`

molobala:
Ohh you're totally right xD !  
I'll change right now

icebob:
Thanks

icebob:
It seems working. I measured transfer times. Streaming 1,5MB files between two processes:
With NATS: ~3000ms
With TCP: ~270ms

Huge difference :)

molobala:
I did not try it with , NATS, I use TCP, not so bad I think :)

icebob:
Thanks. I will still test it, and after we need to cover it with tests.

molobala:
Yes sure :).

zllovesuki:
Sorry but I wanted to ask, was it tested against `back pressure`?

icebob:
I'm thinking how we can add request side streaming. I mean when `ctx.params` is a `Stream`. If we have two side streaming support, we can make services which can encode/decode files, compress/decompress....etc. It will be a big new feature.

molobala:
Yeah really, that will be great, will be better again if we could allow both stream and no stream params together that's to say  `ctx.params` may contains both stream and not stream data.  
If we could do something like that I think that if we could have that possibility  
```js
actions:{
  upload: {
      params: {
         media: {
              type: 'stream'
         },
        extras: {
             type: 'object'
        }
     }
}
```

molobala:
Yeah you're true that what I thought too, right now we can just do like this!  
I think there is no need to make `ctx.params` having both stream and other kind of object, since, in general when uploading file for instance we pass by the gateway , this one will just have to stream the file to the service first then send extra data any.

icebob:
And we have `ctx.meta` to pass other information.

icebob:
Huhh, I think I did it :)

Here is a sample code which is working. It creates an `aes` service which has `encrypt` and `decrypt` action. 
With the `client` broker I open a file, call `aes.encrypt`, the response (which is a stream too) is sent to `aes.decrypt`. And this response (which is a stream again) is saved to a file and checked the result with SHA.
So it send streams on both side, multiple times.

```js
"use strict";

const ServiceBroker = require("../src/service-broker");
const fs = require("fs");
const path = require("path");
const crypto = require("crypto");

const password = "moleculer";

// Create broker #1
const broker1 = new ServiceBroker({
    nodeID: "client-" + process.pid,
    transporter: "NATS",
    logger: console,
    logLevel: "info"
});


// Create broker #2
const broker2 = new ServiceBroker({
    nodeID: "encrypter-" + process.pid,
    transporter: "NATS",
    logger: console,
    logLevel: "info"
});

broker2.createService({
    name: "aes",
    actions: {
        encrypt(ctx) {
            const encrypt = crypto.createCipher("aes-256-ctr", password);

            return ctx.params.pipe(encrypt);
        },

        decrypt(ctx) {
            const decrypt = crypto.createDecipher("aes-256-ctr", password);

            return ctx.params.pipe(decrypt );
        }
    }
});

broker1.Promise.all([broker1.start(), broker2.start()])
    .delay(2000)
    .then(() => {
        broker1.repl();

        const fileName = "d://src.zip";
        const fileName2 = "d://received-src.zip";

        return getSHA(fileName).then(hash1 => {
            broker1.logger.info("Original SHA:", hash1);

            const stream = fs.createReadStream(fileName);

            broker1.call("aes.encrypt", stream)
                .then(stream => broker1.call("aes.decrypt", stream))
                .then(stream => {
                    const s = fs.createWriteStream(fileName2);
                    stream.pipe(s);
                    s.on("close", () => getSHA(fileName2).then(hash => broker1.logger.info("Received SHA:", hash)));
                });
        });

});

function getSHA(fileName) {
    return new Promise((resolve, reject) => {
        let hash = crypto.createHash("sha1");
        let stream = fs.createReadStream(fileName);
        stream.on("error", err => reject(err));
        stream.on("data", chunk => hash.update(chunk));
        stream.on("end", () => resolve(hash.digest("hex")));
    });
}

```

icebob:
If you think, you can test it in [streams](https://github.com/moleculerjs/moleculer/tree/streams) branch.

molobala:
Woww, I'll test it tomrrow !  

That will really be very helpull to me , since  the upload process I use actually are not optimals xD , that will be very good

baadc0de:
I'm also actively using moleculer and faking a few use cases like upload / download / transform with HTTP servers. This feature would really make my day and I could throw away the need to create additional HTTP servers and all the headaches associated with that. Good work! 👍

icebob:
FYI: In 0.13, I improve ProtoBuf & Avro serializers to better handle streaming buffers. They are ~10x faster than JSON serializer in streaming.

Issue #228, opened 2018-03-26
[Proposal/Discussion] Latency-based Balancer


zllovesuki:
My current setup is as follows:
DC A: two `moleculer` services
DC B: two `moleculer` services
DC C: one `moleculer` services
Transporter: gnatsd with TLS on all nodes (each `moleculer` ServiceBroker is connected to its local `gnatsd`)
Persistent data backend: RethinkDB replicated in all DCs

However, DC C is relatively far away (100ms) from A/B (10ms). Currently if I call an action via the web gateway, some of the requests will be routed to DC C even though I'm requesting from DC A.

`db-slugsurvival-*.*` are running on the RethinkDB hosts for database calls, but the rests are scattered and in different environments to be resilient to failures.

# Proposal
Implement some sort of latency-based balancer (like something that Cassandra does). Currently, `moleculer` already has a CPU-based balancer. So we can probably extend on that one somehow...

# Issues
We don't have a reliable way of detecting "latency" if we have a `nats` transporter between all the nodes... I can see how we can determine the latency if we are running `tcp` transporter.

---

If you are interested in my services/actions, here it is...
```
mol $ actions -i -d
╔════════════════════════════════════════════╤═══════════════════════════════════════════════╤══════════╤════════╤═══════════════════════════════════════════════════╗
║ Action                                     │                                         Nodes │ State    │ Cached │ Params                                            ║
╟────────────────────────────────────────────┼───────────────────────────────────────────────┼──────────┼────────┼───────────────────────────────────────────────────╢
║ api-gw.rest                                │                                             5 │    OK    │   No   │                                                   ║
║                                            │                             andromeda-2-17594 │    OK    │        │                                                   ║
║                                            │                             andromeda-0-18568 │    OK    │        │                                                   ║
║                                            │                             andromeda-3-21772 │    OK    │        │                                                   ║
║                                            │                             andromeda-1-55806 │    OK    │        │                                                   ║
║                                            │                              andromeda-5-3436 │    OK    │        │                                                   ║
╟────────────────────────────────────────────┼───────────────────────────────────────────────┼──────────┼────────┼───────────────────────────────────────────────────╢
║ db-slugsurvival-data.fetch                 │                                             5 │    OK    │   No   │ key                                               ║
║                                            │                             andromeda-2-17593 │    OK    │        │                                                   ║
║                                            │                             andromeda-0-18576 │    OK    │        │                                                   ║
[...]

PR #230, opened 2018-03-27
Measure Latency Upon Heartbeat + New Strategy Based on Latency

zllovesuki:
This injects a `transit.sendPing` before each heartbeat, and listen for `$node.pong` event internally to update the node catalog. Similar to CpuUsage but instead of passively taking data, the node will actively probe latency.

This however requires a change in default settings. By default, heartbeat is sent every 5 seconds, which might be too frequent if we are also adding PING/PONG to the mix. Thus, I've also increased the default heartbeatInterval to 10s.

implement #228 

### sidenote
Latency, in theory, could be measured from the passive heartbeat packet. This, however, will require a change in the packet specs (including a `time` property). But the math is too complicated and I don't want to get too much into that... (this assumes time are synced between nodes or we need to do some adjustment which is a pain...)

### roadmap
Maybe in the next major release we will implement passive measurement of latency (instead of active which clots the transporter channel if you have A LOT OF nodes).

### DIFF
--- a/src/registry/node-catalog.js
+++ b/src/registry/node-catalog.js
@@ -44,6 +44,7 @@ class NodeCatalog {
 
 		this.broker.localBus.on("$transporter.connected", this.startHeartbeatTimers.bind(this));
 		this.broker.localBus.on("$transporter.disconnected", this.stopHeartbeatTimers.bind(this));
+		this.broker.localBus.on("$node.pong", this.updateLatency.bind(this));
 	}
 
 	/**
@@ -283,6 +284,20 @@ class NodeCatalog {
 		} else {
 			// Unknow node. Request an INFO from node
 			this.broker.transit.discoverNode(payload.sender);
+		}
+	}
+
+	/**
+	 * Update latency in local node registry
+	 *
+	 * @param {any} payload
+	 * @memberof NodeCatalog
+	 */
+	updateLatency(payload) {
+		const node = this.get(payload.nodeID);
+
+		if (node) {
+			node.updateLatency(payload);
 		}
 	}
 
--- a/src/registry/node.js
+++ b/src/registry/node.js
@@ -37,6 +37,9 @@ class Node {
 		this.rawInfo = null;
 		this.services = [];
 
+		this.historicLatency = [];
+		this.latency = 0;
+
 		this.cpu = null;
 		this.cpuSeq = null;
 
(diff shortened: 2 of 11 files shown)


icebob:
Thanks, but I can't merge it. This implementation is not too efficient. You implemented the latency measuring to the core modules, but I think it is not mandatory. It also gives much overhead to the communication when you don't use the Latency Strategy. E.g. it sends 3 packets instead of 1 packet during a heartbeat cycle. Otherwise, It won't work with TCP transporter because Heartbeat is not implemented in (because it uses Gossip protocol).

I don't think the latency is changed very often. My suggestion is that you should create an individual strategy which doesn't modify the core modules. It starts a timer which sends pings to the nodes, collects the pongs and saves the latencies to the own property. I think the interval can be e.g.: 1 minute. It won't give too much overhead. And better if it is a configurable option.

By the way, is it to collect the last 50 latency and calculate an average is mandatory? Is it not enough to store the last value? Or the last 5 values? If there is something latency issue with some node, the strategy will detect it too slowly (due to the avg). When it detects, maybe the issue is disappeared and the latency returns to the normal value.

_And please don't change the default broker option values._

zllovesuki:
I was thinking about similar things. Just a few things:
1) if I were to send ping in an individual module, what’s the best strategy to access the transit module, and the best strategy to listen for $node.pong?
2) I was already thinking that 50 seems too arbitrary. Maybe I will do a exponential smoothing for the last 10 😂

icebob:
1. If you can wait ~ 1 week I will rewrite `sendPing` to return Promise instead of event triggering. After it, you can send ping via `this.broker.sendPing().then....` 
2. It can be an option. So if somebody don't want to collect, just set it to 1.

zllovesuki:
If I can attach to the local bus then it should be fine for now.

icebob:
You can. Because transit call `broadcastLocal` and it [emit on localBus](https://github.com/moleculerjs/moleculer/blob/master/src/service-broker.js#L1318)

zllovesuki:
Although, `Strategy` can be instantiated multiple times. Thus, my current strategy is to have the first caller to listen an event (`$node.latencyMaster`), and that instance will do the `sendPing`.

Each instance will attach to the localBus for `$node.pong` and dynamically update the lowest latency node. `select` will just return the node.

PR #232, opened 2018-03-27
self-enclosed strategy module

zllovesuki:
per comment from #230, this strategy now does not touch the core module.

implements #228 

Instead,
1) Since `Strategy` can be instantiated multiple times, we will use a "master" to measure latency, and use the localBus to propagate the new latency.
2) Upon first instantiation of `LatencyStrategy`, the caller (usually the registry) will claim `$node.latencyMaster`. It will be responsible for `sendPing` and process `$node.pong` events.
3) It then emits `$node.latencySlave` with node latency data to other instantiations of `LatencyStrategy` (usually are the catalogs), and each instantiation will update its own properties of `nodeLatency`
4) Upon calling `select`, sampling will apply. Code is borrowed from `cpu-usage` but instead of checking against endpoint cpu usage, it will be comparing against the internal list (`nodeLatency`)

still needs better unit test. I'm deploying it on my production system to see if it explodes...

yeah...


icebob:
You should cover better the strategy, sorry.
image

Run `npm test` and open the `coverage/lcov-report/src/strategies/latency.js.html` file to check the coverage report:
image

src/strategies/latency.js
```diff
@@ -0,0 +1,231 @@
...
+		this.mapNode(node);
+
+		let hostMap = this.hostMap.get(node.hostname);
+
+		if (hostMap.historicLatency.length > (this.opts.collectCount - 1))
+			hostMap.historicLatency.shift();
+
+		hostMap.historicLatency.push(payload.elapsedTime);
+
+		avgLatency = hostMap.historicLatency.reduce(function(sum, latency) {
```
icebob: Please use shorthand functions in **all** places without `bind`: `(sum, latency) => {... }`

src/strategies/latency.js
```diff
@@ -0,0 +1,231 @@
...
+			// Get random endpoint
+			if (count == list.length) {
+				ep = list[i];
+			} else {
+				ep = list[random(0, list.length - 1)];
+			}
+			const epLatency = this.hostLatency.get(ep.node.hostname);
+
+			// Check latency of endpoint
+			if (typeof epLatency !== "undefined") {
```
icebob: Please use shorthand check in **all** places: `if (!epLatency) { ... }`.

src/strategies/latency.js
```diff
@@ -0,0 +1,231 @@
...
+			// claim as master
+			this.broker.localBus.on("$node.latencyMaster", function() {});
+			// respond to PONG
+			this.broker.localBus.on("$node.pong", this.processPong.bind(this));
+			// dynamically add new node
+			this.broker.localBus.on("$node.connected", this.addNode.bind(this));
+			// dynamically remove node
+			this.broker.localBus.on("$node.disconnected", this.removeHostMap.bind(this));
+			// try to discovery all nodes on start up
+			this.broker.localBus.on("$broker.started", this.discovery.bind(this));
```
icebob: It is not the best solution, because it is possible that the service starts after the broker started. In this case the timer won't start.
Would be better to call `this.discovery` directly here.

zllovesuki: That's actually okay. If `this.discovery` wasn't called, the module would just behave like RandomStrategy. The "discovery" doesn't have to happen on-time.

zllovesuki:
@icebob The tests were fine. The default parameter with the test timeout (5s) would not test all the situations. Adjusting the parameters to allow all methods to be called within the test timeout gives 100% coverage (without actually changing the code).

src/strategies/latency.js
zllovesuki: \<sidenote>
`setInterval` might potentially cause problems (https://develoger.com/settimeout-vs-setinterval-cff85142555b). Usually, `setTimeout` is a safer invocation. 
\</sidenote>

icebob:
Ok, in this case, we need to use [lolex](https://github.com/sinonjs/lolex) as a fake timer in tests.

src/strategies/latency.js
icebob: OK

README.md
```diff
@@ -1,140 +1 @@
...
-[3rd party licenses](https://app.fossa.io/reports/09fc5b4f-d321-4f68-b859-8c61fe3eb6dc)
-
-# Contact
-Copyright (c) 2016-2018 MoleculerJS
-
-[![@moleculerjs](https://img.shields.io/badge/github-moleculerjs-green.svg)](https://github.com/moleculerjs) [![@MoleculerJS](https://img.shields.io/badge/twitter-MoleculerJS-blue.svg)](https://twitter.com/MoleculerJS)
-
-[paypal]: https://paypal.me/meregnorbert/50usd
-[patreon]: https://www.patreon.com/bePatron?u=6245171
+This is a `bleeding edge` branch of @icebob's project. Upstream is in `upstream` branch. Your warranty is now void.
```
icebob: Please revert the readme.

zllovesuki:
Do you want me to `squash` all those commits?

icebob:
Not necessary

zllovesuki:
Thank you, those code are A LOT more readable 😂

icebob:
Now all codes are covered with unit tests.

icebob:
Would be good to add some integration tests too in the future.

zllovesuki:
As you can probably tell, I’m not the best at writing tests 😂

PR #254, opened 2018-05-02
fix type of ServiceBroker.Promise (fix #251)

### LINKED ISSUE #251: type of Promise field should probably be `typeof Bluebird<any>`
https://github.com/moleculerjs/moleculer/blob/60ddf4736aa29a79a30199c9e7cf5a90d40c120d/index.d.ts#L135

Trying to use this field in my code with `return new this.Promise(...)` makes TS complain about there not being a constructor in the value.

millimoose:


icebob:
@rmccallum81 ?

icebob:
Thanks!

PR #250, opened 2018-04-23
REQUEST FOR COMMENTS: Graceful shutdown of active calls

rmccallum81:
This is an initial step towards a full graceful shutdown of services. Please give comments and ideas for improvements.

PR #255, opened 2018-05-02
REQUEST FOR COMMENTS: Graceful shutdown of active calls

rmccallum81:
Accidental close of #250

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -201,7 +201,7 @@ class Service {
 			return new this.Promise((resolve, reject) => {
 				const timeout = setTimeout(reject, 2000);
 				const checkForContexts = () => {
-					if (this._getActiveContexts().size === 0) {
+					if (this._getActiveContexts().length === 0) {
 						clearTimeout(timeout);
 						resolve();
 					} else {
@@ -289,7 +289,7 @@ class Service {
 		if (_.isFunction(this.broker.ContextFactory.getActiveContexts)) {
 			return this.broker.ContextFactory.getActiveContexts(this);
 		}
-		return new Set();
+		return [];
 	}
 
 	/**
--- a/src/service.js
+++ b/src/service.js
@@ -205,7 +205,7 @@ class Service {
 
 		this.stopped = () => {
 			return new this.Promise((resolve, reject) => {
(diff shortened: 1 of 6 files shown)


icebob:
Please fix tests

src/service.js
```diff
@@ -192,17 +204,30 @@ class Service {
...
-				return this.Promise.method(this.schema.stopped).call(this);
-
-			if (Array.isArray(this.schema.stopped)) {
-				return this.schema.stopped
-					.reverse()
-					.map(fn => this.Promise.method(fn.bind(this)))
-					.reduce((p, fn) => p.then(fn), this.Promise.resolve());
-			}
+			return new this.Promise((resolve, reject) => {
+				const timeout = setTimeout(reject, this.settings.gracefulStopTimeout || this.broker.options.gracefulStopTimeout);
```
icebob: Because it is an "internal" service settings, please follow the conventions and use the `$` prefix. Please change it to `this.settings.$gracefulStopTimeout`

src/service.js
```diff
@@ -90,7 +90,17 @@ class Service {
...
+						if (opts.trackContext) {
+							ctx.dispose();
+						}
+
+						return ret;
+					};
+					const contextDisposeCatch = (ret) => {
+						return this.Promise.reject(contextDispose(ret));
+					};
+					return innerAction.handler(ctx).then(contextDispose, contextDisposeCatch);
```
icebob: Please use the common `.then().catch()` syntax:
```js
return innerAction.handler(ctx).then(contextDispose).catch(contextDisposeCatch);
```

src/service.js
```diff
@@ -260,6 +285,16 @@ class Service {
...
 	}
 
+	/**
+	 * Retrieve list of active contexts for the service
+	 * @returns {Set}
+	 * @private
+	 * @memberof Service
+	 */
+	_getActiveContexts() {
+		return this.activeContexts || [];
```
icebob: Please use `_` before local variable name because it can conflict with user's service variables. Use as `this._activeContexts`

rmccallum81:
I wonder if the transport benchmarks should be updated to include a test with and a test without context tracking. The common benchmark with the FakeTransporter showed a significantly lower reduction in performance.

PR #263, opened 2018-05-09
synchronize node seqs

### LINKED ISSUE #262: dynamically registered services do not update after broker reconnect
When a separate broker reconnects, dynamically registered services (any service that calls .createService after moleculer starts) are not updated because the sequence is out of sync.

I traced the issue here:

https://github.com/moleculerjs/moleculer/commit/fbb99106150d4b41798b6d9550c547f83e9f2aa6#diff-bb81619136bf2c8b0778ee961b076a8cR184

Basically, the sequence # sent in the payload is less than the sequence # stored in the remote broker, ensuring that newly created services are not added to the registry.

fugufish:
on reconnect so dynamically registered service update on remote brokers correctly fixes #262

### DIFF
--- a/src/registry/node-catalog.js
+++ b/src/registry/node-catalog.js
@@ -177,6 +177,7 @@ class NodeCatalog {
 			node.lastHeartbeatTime = Date.now();
 			node.available = true;
 			node.offlineSince = null;
+			node.seq = payload.seq
 		}
 
 		// Update instance
--- a/src/registry/node-catalog.js
+++ b/src/registry/node-catalog.js
@@ -177,7 +177,7 @@ class NodeCatalog {
 			node.lastHeartbeatTime = Date.now();
 			node.available = true;
 			node.offlineSince = null;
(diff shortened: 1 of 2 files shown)


icebob:
Thanks, but I need to test is entirely, because seq value is important in TCP gossip transporter. So maybe this PR solves the problem in other transporters, but I think it can cause mistake in TCP transporter.

fugufish:
let me know if there is anything I can do to help, the issue pretty much breaks any service being able to dynamically load and destroy services and be recognized by a remote broker, so I assume it breaks hot reloading as well in the same scenario.

icebob:
Yes, I know. This bug came with Gossip protocol which brought the `seq`s. Tomorrow I will test & fix it.

icebob:
Thanks. It's fixed in https://github.com/moleculerjs/moleculer/commit/080437d40b8c05d4463044d1eef667b75b4bf2b1

PR #265, opened 2018-05-10
Better differentiation for TCP Service Discovery and UDP Service Discovery.

Wallacy:
Hello,

This was a small change that our team made for TCP/Service Discovery. We are using a UDP only (without gossip) on the last 3 days, and apparently is all working fine.

The rationale is:

- Gossip service discovery is actually a TCP Service Discovery strategy. As the doc point out, is very handy if your network does not support UDP and/or you don't want use. Is very good to a Serviceless node.

- Besides the name "udpDiscovery", the final discovery phase (to interpret the not as available node) is mading using a tcp transport layer.

- processNodeInfo(payload) from node-catalog, is the first responder to change the  "available" status of "discovery nodes" from udp layer (added to offline list) called from "processGossipResponse". At this time,  we actually don't know more from the node compared to the previous state. We only add isReconnected = true and node.lastHeartbeatTime = Date.now() differently that what we can do at the time that addOfflineNode is call.

I add addOnlineNode function and udpCallMeansOnline option to make possible to registrer to make the node avaliable at the time that udp server tell us that this node make a broadcast call.

What other benefits can we expect?
- The gossip protocol can be specialized to use the same transport of the broker and be used as general service discovery for other protocols like WebSockets, Redis, etc.
- UDP can be specialized to be a transport layer in the future ( good for games).

I make 3 videos showing the diference:

Gossip-viz TCP Only:
https://giphy.com/gifs/moleculer-services-gossip-viz-tcp-2tOtQ2l9PhVgcsEi5x

Gossip-viz TCP and UDP (current approach, udpCallMeansOnline  is disable):
https://giphy.com/gifs/gossip-viz-tcp-and-udp-83cOKWP7O5NgAkEFJ2

Gossip-viz UDP Only:
https://giphy.com/gifs/gossip-viz-udo-only-uip5P5TNjM2J99Copz


Wallacy:
Okay, after few more days using this internally, we change the approach a little. I will close this PR and make another one in few days (busy week).

PR #269, opened 2018-05-15
Allow using ES6 class in createService

rmccallum81:


test/unit/service-broker.spec.js
```diff
@@ -1135,6 +1135,15 @@ describe("Test broker.createService", () => {
 		expect(broker.ServiceFactory.mergeSchemas).toHaveBeenCalledWith(schema, mods);
 	});
 
+	it("should load es6 class service", () => {
+		const es6Service = require("../services/greeter.es6._ervice");
+		Object.setPrototypeOf(es6Service, broker.ServiceFactory);
```
icebob: Why need this line?

rmccallum81: Due to the override of the `broker.ServiceFactory` above, it will never detect it as a service class unless the prototype is set to the same object/function. This ensures that the `ServiceFactory` is always the base of the service.

PR #285, opened 2018-06-01
adds ability to customise console object rendering

chrvadala:
## :memo: Description
When trying to log a big object the logged text is written on multiple lines due to the `util.inspect()` behaviour.

Actually `this.logger.info('Test with a big object',  process.release)` prints:

```
[2018-06-01T10:39:20.379Z] INFO  xxx-19680/MATH: Test with a big object { name: 'node',
  lts: 'Carbon',
  sourceUrl: 'https://nodejs.org/download/release/v8.10.0/node-v8.10.0.tar.gz',
  headersUrl: 'https://nodejs.org/download/release/v8.10.0/node-v8.10.0-headers.tar.gz' }
```

With this fix the new log will be:
```
[2018-06-01T10:45:17.710Z] INFO  xxx-19743/MATH: Test with a big object { name: 'node', lts: 'Carbon', sourceUrl: 'https://nodejs.org/download/release/v8.10.0/node-v8.10.0.tar.gz', headersUrl: 'https://nodejs.org/download/release/v8.10.0/node-v8.10.0-headers.tar.gz' }
```

It looks better on console and allows us to work with command line utils like grep, awk...

This fix changes the default `breakLength` to Infinity
> breakLength <number> The length at which an object's keys are split across multiple lines. Set to Infinity to format an object as a single line. Default: 60 for legacy compatibility.
https://nodejs.org/api/util.html#util_util_inspect_object_options

### :gem: Type of change

Add the following code in a service
```
created: {
  this.logger.info('Test with a big object',  process.release)
},
````

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 .DS_Store
 node_modules/
 coverage/
--- a/src/logger.js
+++ b/src/logger.js
@@ -42,9 +42,10 @@ module.exports = {
 	 * @param {Object} bindings
 	 * @param {String?} logLevel
 	 * @param {Function?} logFormatter Custom log formatter function
+	 * @param {Function?} logObjectPrinter Custom object formatter function
 	 * @returns {Object} logger
 	 */
-	createDefaultLogger(baseLogger, bindings, logLevel, logFormatter) {
+	createDefaultLogger(baseLogger, bindings, logLevel, logFormatter, logObjectPrinter) {
 		const noop = function() {};
 
 		const getModuleName = () => {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by icebob** [comment]
Thanks. I'm thinking on it because the multi line print is better at developing. So maybe would be good to change it with a parameter.

**Comment by chrvadala** [comment]
Would you accept a PR on that? If yes, Which could be the name of this flag? 
About the default I think that it should be the inline version.

**Comment by icebob** [comment]
Yes of course. I'm finding a name & place for this option.
I agree that inline should be the default value.

**Comment by icebob** [comment]
Sometimes I have to change the `depth` too in `util.inspect`. So maybe would be good an option which a function, where we can overwrite the object/array printing logic. 
E.g.:
```js
const util = require("util");

    logger: true,
    logLevel: "debug",
    logObjectPrinter: o => util.inspect(o, { depth: 4, colors: false, breakLength: 50 })
});
```
So the `logObjectPrinter` is `null` by default, which means, use the new logic (without breakLength). And if it doesn't fit someone else requirements, they can set a custom function with custom logic.
I think it can cover all options.

icebob:
Could you update the PR, Or do am I?

chrvadala:
No problem, I can work on it. It seems a good improvements for the logging side.
I'll publish an update in few days.

chrvadala:
I added a first version. It should work as expected. LMK

chrvadala:
Great! You are going to document it, right?

icebob:
Yes

chrvadala:
Awesome!

PR #293, opened 2018-06-07
fix metrics d.ts bug

xuming314:
@Action({params:{p1:'string'}, metrics:{params:true}})
async action1(ctx) {...}

@Action({params:{p1:'string', p2:'string'}, metrics:{params:['p1', 'p2']}})
async action2(ctx) {...}

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -26,7 +26,7 @@ declare namespace Moleculer {
 	type ActionParams = { [key: string]: ActionParamTypes };
 	type MetricsParamsFuncType= (params: object) => object;
 	type MetricsMetaFuncType= (meta: object) => object;
-	type MetricsOptions = { params?: boolean | string[] | MetricsParamsFuncType , meta?: boolean | string[] | MetricsMetaFuncType };
+	type MetricsOptions = { params?: boolean | string[] | MetricsParamsFuncType, meta?: boolean | string[] | MetricsMetaFuncType };
 
 	interface Action {
 		name: string;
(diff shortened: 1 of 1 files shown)


icebob:
@rmccallum81 please review it. Thanks

index.d.ts
```diff
@@ -24,7 +24,7 @@ declare namespace Moleculer {
 	type ActionParamSchema = { [key: string]: any };
 	type ActionParamTypes = "boolean" | "number" | "string" | "object" | "array" | ActionParamSchema;
 	type ActionParams = { [key: string]: ActionParamTypes };
-	type MetricsOptions = { params?: "boolean" | "array" | "function", meta?: "boolean" | "array" | "function" };
+	type MetricsOptions = { params?: boolean | string[] | function, meta?: boolean | string[] | function };
```
rmccallum81: `function` isn't a type, please use a function fingerprint to indicate input parameters and the output of the function.

Is the array only ever going to be an array of strings or could it be numbers or objects?

@icebob perhaps you could clarify the expected function parameters if any?

icebob: The array can be only string array.
The function signature is `function(params: ActionParams)` or `function(meta: object)`

rmccallum81: The functions have no return type (meaning void)?

icebob: Ohh sorry. But it can return `any` :)

icebob:
@rmccallum81 could you finish this PR, please? As I see, xuming314 couldn't finish it and tomorrow I will make a 0.12 release.

Thanks.

rmccallum81:
I've created a new PR with the changes #299

PR #324, opened 2018-07-11
Different TS version did not error but gave warning

rmccallum81:
I didn't realize that the TS tests were not part of the standard tests. I had created my own service separately to test which had a different TS version and config which was less strict. I've run the full set of tests with this and they pass.


icebob:
Thanks!

Issue #273, opened 2018-05-18
AMQP transporter issue


katsanva:
Hello.

Even in idle state, some messages are not removed from the queue that causes memory to go out of the bounds.

![Alt text](https://monosnap.com/image/5BjSbb4vlqvAsNRV5MXNt3IhsCY58e)

I see that on average 3 messages are sent to the queue per second. On the other hand, no application is consuming the messages.

It seems that amqp transporter makes a "memory leak" somehow. 

@icebob Do you have any thoughts on that? Is there any configuration to prevent that?

icebob:
RabbitMQ stores messages while they are not consumed. I think it is not a transporter issue, it is a configuration problem, but I don't know very well RabbitMQ.

Nathan-Schwartz:
@katsanva It sounds like the issue isn't memory, but that you weren't expecting AMQP to store your messages. In the screenshot you shared it looks like only 1/3rd of the memory is being used.

AMQP will store your messages until you process or purge them. This is an intended behavior to make message delivery more reliable.

If this behavior isn't helpful for your use case, it may be best to switch to a different transporter like NATS :)

Nathan-Schwartz:
Ahh, I see what you mean @katsanva . I thought you were talking about requests and such. I agree that ping-pongs and hearbeats (etc) should be discarded.

The current solution is to automatically delete those queues whenever the consumer disconnects. Since each node would have its own HEARTBEAT queue, I would expect them to be deleted as soon as a broker disconnects, so those messages wouldn't pile up.

What packet was associated with the queue in your screenshot?

Have you had issues with HEARTBEATS, PING, PONG, DISCOVER, DISCONNECT, or INFO packets piling up?



If there is a case not reasonably covered by auto-deletion, a default expiration time could be set by setting a default message timeToLive here: https://github.com/moleculerjs/moleculer/blob/master/src/transporters/amqp.js#L244

That would do the trick I think, though it is possible to run into a little weirdness because of this behavior:
> Only when expired messages reach the head of a queue will they actually be discarded (or dead-lettered). Consumers will not have expired messages delivered to them.
- https://www.rabbitmq.com/ttl.html#per-message-ttl-caveats

An alternative approach would be to set a policy to enforce a queue limit for those queues (haven't tried this).

Even though there are a lot of HEARTBEAT/etc messages enqueued, my experience has been that brokers process them extremely quickly, and it hasn't been an issue for me in the past.

katsanva:
[...]

PR #341, opened 2018-07-27
Add option to enable auto delete queues in amqp transport

urossmolnik:
## :memo: Description

In some systems, it is impossible to generate constant unique nodeIDs (ex. multiple instances of Heroku dynos). In this case nodeID must has some random part.  
Problem with doing this and using AMQP transport is you end up with a lot of obsolete queues in RabbitMQ.  

That's why I added `autoDeleteQueues` option to amqp options. If true queues will be autodeleted once service is stopped (queue listener removed).  
Of course you loose some functionality in this case. Once service is re-started it will have new request queue, meaning not finished reqeusts will be lost.  

### :gem: Type of change

### :scroll: Example code
```js
  ...
  transporter: {
    type: 'AMQP',
    options: {
      ...
      autoDeleteQueues: true,
    }
  },
  ...
})
``` 

If new option is not set funcionality is not changed.  
Manually tested that when option is enabled, services still communicate successfully and queues get deleted after service is stop.  

### DIFF
--- a/src/transporters/amqp.js
+++ b/src/transporters/amqp.js
@@ -73,6 +73,12 @@ class AmqpTransporter extends Transporter {
 		if (typeof opts.consumeOptions !== "object")
 			opts.consumeOptions = {};
 
+		opts.autoDeleteQueues =
+			opts.autoDeleteQueues === true ? 2*60*1000 :
+				typeof opts.autoDeleteQueues === "number" ? opts.autoDeleteQueues :
+					opts.autoDeleteQueues === false ? -1 :
+						-1; // Eventually we could change default
+
 		super(opts);
 
 		this.hasBuiltInBalancer = true;
@@ -216,13 +222,13 @@ class AmqpTransporter extends Transporter {
 		switch(packetType) {
 			// Requests and responses don't expire.
 			case PACKET_REQUEST:
-				packetOptions = this.opts.autoDeleteQueues === true && !balancedQueue
-					? { autoDelete: true }
+				packetOptions = this.opts.autoDeleteQueues >= 0 && !balancedQueue
+					? { expires: this.opts.autoDeleteQueues }
 					: {};
 				break;
 			case PACKET_RESPONSE:
-				packetOptions = this.opts.autoDeleteQueues === true
-					? { autoDelete: true }
+				packetOptions = this.opts.autoDeleteQueues >= 0
+					? { expires: this.opts.autoDeleteQueues }
 					: {};
 				break;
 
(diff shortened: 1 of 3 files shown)


urossmolnik:
Or am I missing something and this can be achieved in some other way?

Nathan-Schwartz:
@urossmolnik Thanks for investing time into the AMQP transporter!

I think auto-deleting the node specific queues (heartbeats, etc) is a good idea since they aren't intended to be shared or reused. I believe most of the node-specific queues are already set to auto-delete and some have message TTL enabled, though I'm not sure if that has been working as expected (#273 may be related). I think this case where most of those obsolete queues would be generated.

It has been a little while since diving into this code, but I think this solution could be problematic for at least one edge case as it is currently implemented. The reason is that the load balanced methods don't use the nodeID in their topic names so that queues are shared between all nodes and requests can be balanced between them. The only time a load balanced request queue name should become obsolete is if there is a code change that alters the action or service name (please correct me if I'm wrong).

I think in the case where the queue has messages that must be delivered and all the services go down for a moment due to a power outage or daily Heroku restart, the autodelete functionality would trigger and those messages would be lost.

A general solution I've used in the past is [Queue TTL policy](https://www.rabbitmq.com/ttl.html#queue-ttl). It has its own downsides but it felt a little safer to me.

urossmolnik:
@Nathan-Schwartz thanks for feedback.

I didn't have any problems with HEARTBEAT, ... queues. They were always auto-deleted for me.

I had problems with the rest of the queues having nodeID in their name (REQ and RES). These are auto-deleted now if you specify so in options.  
Queues used by balanced methods are still **not deleted**, even if `autoDeleteQueues` option is on.

Nevertheless I like your Queue TTL policy idea as a safer option 👍. I'll implement that in next days and update this pull request.

urossmolnik:
@Nathan-Schwartz no I wasn't suggesting building TTL policy into the transporter. I was just contemplating if I should set expire to balanced queues as well (if `autoDeleteQueues` is true). But I figured out I can't do this since different services could try asserting same queue with different options.  

I don't think we can make this option default to true. It would be a breaking change. Asserting existing queue with different options would fail.

Nathan-Schwartz:
It would be a breaking change, but I think expiring or auto-deleting all node specific queues is a good default behavior. Do you agree? The change would need to be documented in the next release, and maybe wait for v0.14 to roll it out though.

Regarding expiring balanced queues, I don't think we should do it by default and if it is supported it might be nice to make the expiration time configurable

urossmolnik:
I decided not to make it default. But it can be changed easily in future.  

```js
opts.autoDeleteQueues =
  opts.autoDeleteQueues === true ? 2*60*1000 :
    typeof opts.autoDeleteQueues === "number" ? opts.autoDeleteQueues :
      opts.autoDeleteQueues === false ? -1 :
        -1; // Eventually we could change default
```  

As you see queues will auto-delete (expire) if `autoDeleteQueues` is true or a positive number (including zero). I chose 2 minutes expire time as default.

icebob:
Thanks @urossmolnik for the PR and thanks for @Nathan-Schwartz for reviewing. I read the comments and I'm confused. So the default value is `-1`, which means it is not a breaking change and can be merged?

urossmolnik:
Yes @icebob, the default value is `-1` and it means there are no breaking changes.

omerts:
This doesn't work for all queues, some of the queues don't get this value:
image

This is my config:
image

Issue #337, opened 2018-07-23
TypeError occured when es6 service class ran thourgh the moleculer-runner cli


zerocool-kor:
I'm using **moleculer 13.1**, and writing some service class using ES6 class type. 
It's OK and there is no problem when using that service class in `broker.loadService` API, 
But service class can not be loaded through the moleculer-runner cli.

### es6 service class source code (device-gateway-mqtt-server.service.js)
```javascript
const Service = require("moleculer").Service;

class DeviceGatewayMQTTServer extends Service {
  constructor(broker) {
    super(broker);

    this.parseServiceSchema({
      name: "device-gateway-mqtt-server",
      created: this.serviceCreated,
      started: this.serviceStarted,
      stopped: this.serviceStopped,
    });
  }

  serviceCreated() {
    this.logger.info("ES6 Service created.");
  }

  serviceStarted() {
    this.logger.info("ES6 Service started.");
  }

  serviceStopped() {
    this.logger.info("ES6 Service stopped.");
  }
}

module.exports = DeviceGatewayMQTTServer;

```

### No proplem, when using the broker.loadService api 
```javascript
const { ServiceBroker } = require('moleculer');
const broker = new ServiceBroker(require('./config/moleculer.config.js'));
broker.loadService('./src/device-gateway-mqtt-server.service.js');
broker.start();
```
```
$ node ./test.js
[2018-07-23T01:43:42.199Z] INFO  kmc-yj-d004733-6384/BROKER: Moleculer v0.13.1 is starting...
[2018-07-23T01:43:42.199Z] INFO  kmc-yj-d004733-6384/BROKER: Node ID: kmc-yj-d004733-6384
[2018-07-23T01:43:42.199Z] INFO  kmc-yj-d004733-6384/BROKER: Namespace: s**********
[2018-07-23T01:43:42.199Z] INFO  kmc-yj-d004733-6384/REGISTRY: Strategy: RoundRobinStrategy
[2018-07-23T01:43:42.215Z] INFO  kmc-yj-d004733-6384/BROKER: Serializer: JSONSerializer
[2018-07-23T01:43:42.215Z] INFO  kmc-yj-d004733-6384/BROKER: Transporter: TcpTransporter
[2018-07-23T01:43:42.215Z] INFO  kmc-yj-d004733-6384/BROKER: Registered 10 internal middleware(s).
[...]

Issue #328, opened 2018-07-17
Load service as es6 module


rhrn:
* api.service.js
```js
import ApiService from 'moleculer-web'

export default {
  name: 'test',
  mixins: [ApiService],
  actions: {
    hello () {
      return 'Hello API Gateway!'
    }
  }
}
```
It's exports as `exports.default`

One of solution override schema by schema.default
```js
if (schema.default) {
  schema = schema.default
}
// schema = schema.default || schema
```
at https://github.com/moleculerjs/moleculer/blob/v0.13.1/src/service-broker.js#L533

icebob:
Could you create a PR with tests?

Wallacy:
Also this can help:
https://www.npmjs.com/package/esm

If used on service-broker ES6 classes should work.

tiaod:
Why this got closed?

cs4inctec:
con confirm that this is still not working.

piranna:
In moleculer 0.14 is still happening. With Typescript default configuration it gets transpiled to CommonJS, but later it can't load EcmaScript Modules dependencies, and when changing its configuration to generate ESM modules I get the next error:

```js
 Failed to load service '/home/piranna/Trabajo/Dyte/dyte-remote-router-provider/dist/services/bootstrap.service.js' Error [ERR_REQUIRE_ESM]: require() of ES Module /home/piranna/Trabajo/Dyte/dyte-remote-router-provider/dist/services/bootstrap.service.js from /home/piranna/Trabajo/Dyte/dyte-remote-router-provider/node_modules/moleculer/src/service-broker.js not supported.
Instead change the require of bootstrap.service.js in /home/piranna/Trabajo/Dyte/dyte-remote-router-provider/node_modules/moleculer/src/service-broker.js to a dynamic import() which is available in all CommonJS modules.
    at ServiceBroker.loadService (/home/piranna/Trabajo/Dyte/dyte-remote-router-provider/node_modules/moleculer/src/service-broker.js:788:14)
    at /home/piranna/Trabajo/Dyte/dyte-remote-router-provider/node_modules/moleculer/src/runner.js:413:50
    at Array.forEach (<anonymous>)
    at MoleculerRunner.loadServices (/home/piranna/Trabajo/Dyte/dyte-remote-router-provider/node_modules/moleculer/src/runner.js:413:25)
[...]

PR #343, opened 2018-07-27
Load es6 modules and changing service schema load checker.

Wallacy:
### :gem: Type of change

### :scroll: Comments 

This is linked to #337 and #328

I came up with this problem in my personal project, and that was the solution I found (with the @rhrn tip)

- [x] Full: service-broker.spec
- [x] Added: "should load es6 module class service"
- [x] Changed: "should load 6 services"

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2496,6 +2496,11 @@
       "integrity": "sha512-qzm/XxIbxm/FHyH341ZrbnMUpe+5Bocte9xkmFMzPMjRaZMcXww+MpBptFvtU+79L362nqiLhekCxCxDPaUMBQ==",
       "dev": true
     },
+    "esm": {
+      "version": "3.0.72",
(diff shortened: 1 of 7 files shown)


Wallacy:
For some reason travis-ci does not run the updated test.

icebob:
Thanks for the PR, nevertheless, I don't think so that `esm` is necessary for Moleculer. Currently the `schema.default` checking is enough.

Wallacy:
I think jest is not fully compatible with esm yet! jest.mock("../../../src/transporters/tcp/tcp-reader") is producing another module without all required properties for esm.

Wallacy:
Without esm (or babel) moleculer will not load es6 modules because require will not parse "import * from Y".

https://nodejs.org/api/esm.html

Anyway, i think will be more wise to use babel directly instead esm.

Wallacy:
I will make another PR to handle the problem separately.

icebob:
OK

Issue #321, opened 2018-07-10
Can you change 'HEARTBEAT' log level from 'debug' to 'trace' ?


roytan883:
The `HEARTBEAT` logs always flush `debug` logs, can you move it to `trace` or anywhere i can disable `HEARTBEAT` logs ?

```
18-07-10_11:31:14.282 DEBUG transit: Incoming EVENT packet from 'rts-auth-dev-88-0'
18-07-10_11:31:14.286 DEBUG transit: Event 'metrics.trace.span.finish' received from 'rts-auth-dev-88-0' node in 'metrics' group(s).
18-07-10_11:31:14.287 DEBUG transit: Incoming EVENT packet from 'rts-demo-dev-88-1'
18-07-10_11:31:14.288 DEBUG transit: Event 'metrics.trace.span.finish' received from 'rts-demo-dev-88-1' node in 'metrics' group(s).
18-07-10_11:31:14.289 DEBUG transit: Incoming EVENT packet from 'rts-demo-dev-88-1'
18-07-10_11:31:14.290 DEBUG transit: Event 'metrics.trace.span.finish' received from 'rts-demo-dev-88-1' node in 'metrics' group(s).
18-07-10_11:31:14.290 DEBUG transit: Incoming EVENT packet from 'rts-apiGateway-dev-88-0'
18-07-10_11:31:14.290 DEBUG transit: Event 'metrics.trace.span.finish' received from 'rts-apiGateway-dev-88-0' node in 'metrics' group(s).
18-07-10_11:31:14.504 DEBUG transit: Incoming HEARTBEAT packet from 'rts-apiGateway-dev-88-0'
18-07-10_11:31:14.706 DEBUG transit: Incoming HEARTBEAT packet from 'rts-demo-dev-88-0'
18-07-10_11:31:15.102 DEBUG metrics: Span sending error! Cannot decode spans [object Object],[object Object],[object Object],[object Object]
18-07-10_11:31:17.129 DEBUG transit: Incoming HEARTBEAT packet from 'rts-auth-dev-88-0'
18-07-10_11:31:17.288 DEBUG transit: Send HEARTBEAT packet to '<all nodes>'
18-07-10_11:31:17.559 DEBUG transit: Incoming HEARTBEAT packet from 'rts-demo-dev-88-1'
18-07-10_11:31:19.505 DEBUG transit: Incoming HEARTBEAT packet from 'rts-apiGateway-dev-88-0'
[...]

PR #352, opened 2018-08-04
Changed debug into trace for packet logs

WoLfulus:
## :memo: Description

Related to #321


icebob:
The `console.trace` prints stack traces too for every message... It's not too good.
image
Maybe need to change the logger that it uses `console.debug` for `trace` messages too.

WoLfulus:
I'll take a better look at this ASAP

ryanm-pm:
A verbosity setting could be a good solution to this. Leave it in the debug level but add a verbosity input to the logger function with a default of 1. If the level is set higher in the broker settings, include those of a higher number.

WoLfulus:
i havent had time to take a proper look at this again :(

icebob:
Other PR will be merged #373 instead of this. Sorry.

Issue #362, opened 2018-08-15
Unhandled rejection TypeError: this.heartbeatTimer.unref is not a function


fugufish:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [ x] I am running the latest version
- [ x] I checked the documentation and found no answer
- [ x] I checked to make sure that this issue has not already been filed
- [x ] I'm reporting the issue to the correct repository

## Failure Information
When using any transporter but the default within my specs I receive the following:
```
Unhandled rejection TypeError: this.heartbeatTimer.unref is not a function
        at NodeCatalog.startHeartbeatTimers (/Users/fugufish/Code/lucid/node_modules/moleculer/src/registry/node-catalog.js:63:23)
        at EventEmitter.Object.<anonymous>.EventEmitter.emit (/Users/fugufish/Code/lucid/node_modules/eventemitter2/lib/eventemitter2.js:384:22)
        at ServiceBroker.broadcastLocal (/Users/fugufish/Code/lucid/node_modules/moleculer/src/service-broker.js:1113:18)
        at Promise.resolve.then.then.delay.then (/Users/fugufish/Code/lucid/node_modules/moleculer/src/transit.js:91:17)
        at tryCatcher (/Users/fugufish/Code/lucid/node_modules/bluebird/js/release/util.js:16:23)
        at Promise._settlePromiseFromHandler
```

It does not occur when running the app through moleculer runner. While debugging I noted that heartbeatTimer is not a Timer but a number.

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.13.1 ( behavior was present in 0.12.8)
* NodeJS version: 10.4.1
* Operating System: Mac

icebob:
How do you debug it?

fugufish:
breakpoints + webstorm. This test will throw it if you set the broker's transporter to anything but default.

```
https://github.com/lucid-mud/engine/blob/develop/src/World/Services/DataServices/PlayerDataService/PlayerDataService.spec.ts
```

fugufish:
No. And I don't see anything like it in my package lock. Like I said in my original post, the object is not a Timer, but a number. After some investigating it seems that this is a behavior of jest mocking the browser, the setting `"testEnvironment": "node"` solves the issue.

PR #364, opened 2018-08-17
Fail fast for jest when "testEnvironment": "node" is not set

fugufish:
## :memo: Description
Fail fast when `hearbeatTimer` is a number. This occurs in scenarios when a developer is using Jest for testing and they have failed to set `"testEnvironment": "node"`. By throwing an error here it save development time troubleshooting.

### :dart: Relevant issues
ref: #362 

### :gem: Type of change

- [x ] New feature (non-breaking change which adds functionality)

Write a simple Jest test that does't override the default `testEnvironment` configuration and attempt to start a broker. An error will throw referencing issue #362 which has enough information to fix their code.

### DIFF
--- a/src/registry/node-catalog.js
+++ b/src/registry/node-catalog.js
@@ -61,7 +61,7 @@ class NodeCatalog {
 
 		}, this.broker.options.heartbeatInterval * 1000);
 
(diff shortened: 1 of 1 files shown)


fugufish:
there is no clear way to test this particular issue without using manual mock of setTimeout, which would likely break other tests.

Issue #387, opened 2018-09-27
Unnecessary re-connection attempts when working with multiple servers


Gadi-Manor:
I'm working with multiple servers (master-slave) and I came across the following issue:
Assuming I have two broker objects:
```js
GoodBroker = {
        host: 'exist.com',
        protocol: 'mqtts',
        port: 8883
      }
BadBroker = {
        host: 'notExist.com',
        protocol: 'mqtts',
        port: 8883
      }
```

I setup the transporter object like this:
```js
transporter = {
    type: "MQTT",
    options: {
       username: 'username',
       password: 'password',
       servers: [firstBroker, secondBroker]
   }
}
```
Now, what happens is:
1. firstBroker = GoodBroker, secondBroker = BadBroker: Connecting to the first broker as expected
2. firstBroker = BadBroker, secondBroker = GoodBroker: Failed to connect to the first, then successfully connected to the second. So far so good. However at that point it endlessly attempting to connect to the first broker again.

I saw that in src/transit.js there is a a part of code that calls connect again every time it fails. I think this is unnecessary. When I removed it it worked as expected.

It's important to say that I checked to see if maybe that part is important when working with a single broker and I found that from what I can tell it's not. 

I changed the configuration to work with a single broker which was down. When I run the application I got (again, at this point I removed the re-connection):
```js
WARN  - MQTT client is reconnecting...
WARN  - MQTT client is disconnected.
WARN  - MQTT client is reconnecting...
WARN  - MQTT client is disconnected.
WARN  - MQTT client is reconnecting...
// and a lot more
```
Then, I started the broker and it was connected immediately. I think the mqtt module already handles the re-connection mechanism but since I'm new to molecular I wanted to know what you think before I PR.

icebob:
Interesting. Please attach a full log from the start. The reconnecting part is important for all other transporters, so we can't remove it. Would be better to handle this situation in the MQTT transporter code, if it's available.

Gadi-Manor:
INFO  - Moleculer v0.13.2 is starting...
INFO  - Node ID: server-1
INFO  - Namespace: ns
INFO  - Strategy: RoundRobinStrategy
INFO  - Serializer: JSONSerializer
INFO  - Transporter: MqttTransporter
INFO  - Registered 10 internal middleware(s).
INFO  - Connecting to the transporter...
ERROR - MQTT error. connect ECONNREFUSED 34.249.88.38:8883
WARN  - MQTT client is disconnected.
[...]

PR #391, opened 2018-10-09
Fix issue #387

Gadi-Manor:
Unnecessary re-connection attempts when working with multiple servers #387

## :memo: Description
I'm working with multiple servers (master-slave) and I came across the following issue:
Assuming I have two broker objects:
```js
GoodBroker = {
        host: 'exist.com',
        protocol: 'mqtts',
        port: 8883
      }
BadBroker = {
        host: 'notExist.com',
        protocol: 'mqtts',
        port: 8883
      }
```

I setup the transporter object like this:
```js
transporter = {
    type: "MQTT",
    options: {
       username: 'username',
       password: 'password',
       servers: [firstBroker, secondBroker]
   }
}
```
Now, what happens is:
1. firstBroker = GoodBroker, secondBroker = BadBroker: Connecting to the first broker as expected
2. firstBroker = BadBroker, secondBroker = GoodBroker: Failed to connect to the first, then successfully connected to the second. So far so good. However at that point it endlessly attempting to connect to the first broker again.

### :dart: Relevant issues

#387

### :gem: Type of change

### :scroll: Example code
```js

``` 

I saw that in src/transit.js there is a a part of code that calls connect again every time it fails. I think this is unnecessary. When I removed it it worked as expected.

It's important to say that I checked to see if maybe that part is important when working with a single broker and I found that from what I can tell it's not. 

To test it I changed the configuration to work with a single broker which was down. When I run the application I got (again, at this point I removed the re-connection):
```js
WARN  - MQTT client is reconnecting...
WARN  - MQTT client is disconnected.
WARN  - MQTT client is reconnecting...
WARN  - MQTT client is disconnected.
WARN  - MQTT client is reconnecting...
// and a lot more
```
Then, I started the broker and it was connected immediately.


src/transit.js
```diff
@@ -115,10 +115,12 @@ class Transit {
 					this.logger.warn("Connection is failed.", err.message);
 					this.logger.debug(err);
 
-					setTimeout(() => {
-						this.logger.info("Reconnecting...");
-						doConnect();
-					}, 5 * 1000);
+					if (this.tx.opts && this.tx.opts.protocolId !== "MQTT") {
```
icebob: I think it doesn't filter out that there is one or multiple servers configured, right?

Gadi-Manor: I'm not sure I understood you and if not please explain to me again.

You are right, it only filter out by protocol.
The re-connection is not an issue when there is one server configured. 
On cases of multiple servers configured there could be unnecessary re-connection attempts.

However, since the mqtt lib already handles the re-connection mechanism we can comment out the re-connection from here for MQTT protocol only (regardless of number of servers). 

Therefore it doesn't matter how many servers are configured, if we are dealing with MQTT we should leave the re-connection to the mqtt lib.

ryanm-pm: A flag on the `tx.opts` (like `noReconnect`) object would be better suited since this could end up being applied to other protocols later on. The transporter constructor could set a default for this but also allow for it to be passed in as this would allow for more customization when creating your own transporters.

Gadi-Manor:
Updated

Gadi-Manor:
So to disable it you should set `transit: { disableReconnect = true }` when creating a service broker?
If so  I think it's a little unclear. It looks like there is no re-connection at all when in fact it only leaves the it to the transport (at least that's the case for MQTT).

icebob:
Thanks!

PR #398, opened 2018-10-23
AMQP connection recovery + Support for AMQP cluster connection

urossmolnik:
## :memo: Description

**AMQP connection recovery fix**  

Connection to AMQP was not recovering - rejecting resolved promise doesn't work.  
Am I missing something? Should service not automatically recover once connected? Which brings a question, does reconnect work for other transits?  
I resolved this on AMQP transit by passing error callback to connect method and calling this error callback on errors. 

**Support for AMQP cluster connections/mulitple HA proxies**

AMQP `url` can now be an array of URLs or semicolon delimited string containing multiple URLs.  
Nothing fancy, just round robin try thru urls when re-connecting.  

### :gem: Type of change

**AMQP connection recovery**  

No automatic tests added. AMQP connection successfully recovered after restarting RabbitMQ service.

**Support for AMQP cluster connections/multiple HAproxies**

All existing tests pass (covering existing use case with single url).  
Running with multiple urls in production.  

### DIFF
--- a/src/transit.js
+++ b/src/transit.js
@@ -108,18 +108,22 @@ class Transit {
 			this.__connectResolve = resolve;
 
 			const doConnect = () => {
-				/* istanbul ignore next */
-				this.tx.connect().catch(err => {
+				let reconnectStarted = false;
+				const errorHandler = (err) => {
 					if (this.disconnecting) return;
+					if (reconnectStarted) return;
 
-					this.logger.warn("Connection is failed.", err.message);
+					this.logger.warn("Connection is failed.", err && err.message || 'Unknown error');
 					this.logger.debug(err);
 
+					reconnectStarted = true;
 					setTimeout(() => {
 						this.logger.info("Reconnecting...");
 						doConnect();
 					}, 5 * 1000);
-				});
+				};
+				/* istanbul ignore next */
+				this.tx.connect(errorHandler).catch(errorHandler);
 			};
 
 			doConnect();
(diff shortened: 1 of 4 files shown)


Nathan-Schwartz:
Hey @urossmolnik thanks for tackling this. Yes, transporters should definitely reconnect and I believe all of the others do it. You can read through the comments on #72 if you are curious why it wasn't implemented initially.

I think the other transporters rely on the client libraries to handle reconnection, and if we wanted to go that route we might be able to use [amqp-connection-manager](https://www.npmjs.com/package/amqp-connection-manager). I've used it recently and so far have had positive experiences.

Nathan-Schwartz:
Sorry I should have sent you a more exact link. 

Yes, that comment and the following conversation are what I was referring to.

>I handle reconnect in transit so the whole amqp connect logic will go thru again. Meaning all exchanges/queues will be created/bounded again.
Maybe I am missing some cleanup that should be done (and that would bring too much complexity)?

No, I think that reconnecting from transit cares for the cases I was concerned about. I guess I was so focused on making the transporter match the other implementations that I didn't think of it. 🙂 

It seems like a good solution, so long as @icebob doesn't have any concerns about how it might affect other transporters. 👍

urossmolnik:
@icebob any thoughts about this solution?

icebob:
@urossmolnik thank you for your PR. I'm not enough familiar with AMQP, so if @Nathan-Schwartz said your PR is good, I will merge it immediately :)

urossmolnik:
@icebob question for you is to double check changes in transit.js :)  
https://github.com/moleculerjs/moleculer/pull/398/commits/3da39a4f03f965bfa3be3d3f5fd6da6b34d289de  

`connect` in transit.js is passing `errorHandler` function to `connect` in transporter. Other transporters are just ignoring this param so nothing should be broken.

Issue #409, opened 2018-11-05
this.broker.cacher.clean does not accept arrays


dkuida:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

## Expected Behavior

based on the documentation I should be able to call
```
this.broker.cacher.clean([ "users.**", "posts.**" ]);
```
as an array 
same goes for 

this.broker.cacher.del([ "users.**", "posts.**" ]);

but with redis cacher, 

the clean function starts with 

```
clean(match = "*") {
		return new Promise((resolve, reject) => {
			match = this.prefix + match.replace(/\*\*/g, "*");

````
and fails on the match as it expects a string.

* moleculer 0.13:
* node 10

PR #411, opened 2018-11-08
Hotfix/cache clear arrays

dkuida:
## :memo: Description
fix of BUG #409

* utilization of the built in scanStream of ioredis, rather than manual command with callbacks.
* rewrite of the callback structure to avoid using 
```javascript
let self = this;
```
with arrow functions

several unit tests, as well as manual testing against real redis DB

- [ ] Test A
- [ ] Test B

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [x ] **I have added tests that prove my fix is effective or that my feature works**
- [x ] **New and existing unit tests pass locally with my changes**

### DIFF
--- a/dev/dev.js
+++ b/dev/dev.js
@@ -1,5 +1,4 @@
 "use strict";
-
 const ServiceBroker = require("../src/service-broker");
 const util = require("util");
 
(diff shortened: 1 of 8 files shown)


dkuida:
@icebob will do - but looking at ti further - node 6 reaches end of life april 2019, what is you take up until which point to support it?

icebob:
I will support Node 6 till April 2019. By the way, the Promise has better performance than async/await, so I don't plan to rewrite the whole codebase from Promise to async/await. Better performance comes only in Node 10 (maybe). So it means after April, we can use async/await in tests but not in the performance-critical functions in the core.

So please change this PR code to Promise. Thanks.

PR #412, opened 2018-11-08
deleting the workaround because the bug was fixed in version v1.0.1

caiobessa:
## :memo: Description

I deleted the workaround in the code because the bug was closed in the folowing issue https://github.com/nats-io/node-nats/issues/229 

### :dart: Relevant issues
 Don't have any


icebob:
Thank you!

PR #414, opened 2018-11-12
lint

dkuida:
## :memo: Description

no functionality - only lint cleaning to enable lint rules check on every PR

### DIFF
--- a/src/serializers/base.js
+++ b/src/serializers/base.js
@@ -20,7 +20,7 @@ class Serializer {
 	 *
 	 * @memberof Serializer
 	 */
-	constructor() {
+	constructor () {
 	}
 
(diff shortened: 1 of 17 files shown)


src/serializers/base.js
```diff
@@ -30,7 +30,7 @@ class Serializer {
 	 *
 	 * @memberof Serializer
 	 */
-	init(broker) {
+	init (broker) {
```
icebob: What kind of tool did you use for formatting? Because I don't like this form that it adds extra space between method name and params.

icebob:
Sorry, but I don't merge it, because I've added some new ESList rules which conflict this changes.

PR #420, opened 2018-11-18
Hotfix/linting

dkuida:
All linting now passing

Could you consider to set linting as a precondition for pull requests and pushes?


icebob:
How can we set/check it?

PR #428, opened 2018-12-07
service typing should be consistent in naming.

fugufish:
## :memo: Typescript typing should ideally have consistent naming

Rename `Actions` typing to `ServiceActions` to reflect consistent naming conventions

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -65,6 +65,7 @@ declare namespace Moleculer {
 	}
 
 	type ServiceActions = { [key: string]: Action | ActionHandler; };
+	type Actions = ServiceActions;
 
 
 	class Context<P = GenericObject, M = GenericObject> {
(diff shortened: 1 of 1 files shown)


icebob:
It can cause breaking change in Typescript projects? @rmccallum81 ?

fugufish:
Yes it could. Backwards compatibility could be maintained if we exported
both definitions. That said having inconsistent naming becomes really
confusing and larger typescript projects

icebob:
@fugufish please update PR with alias.

AndreMaz:
I think that [visibility](https://moleculer.services/docs/0.13/actions.html#Action-visibility) param is missing in Action definition.

icebob:
Somebody can fix this PR and add visibility? At this weekend I will release 0.13.5 and it can contain this PR, as well.

AndreMaz:
It would be great if someone with good knowledge of typescript could also solve this https://github.com/moleculerjs/moleculer-template-project-typescript/issues/7

fugufish:
@icebob should be fixed now. Actions and ServiceActions types should be interchangeable as Actions now aliases Actions

PR #442, opened 2019-01-04
Fix incompatibility issue between ES6 Promise and Bluebird in Typescript definitions

snnsnn:
## :memo: Description
I added async started and stopped functions to my service schema but Typescript gives the error below:

```js
import http from 'http';
import { ServiceSchema } from 'moleculer';

const serviceSchema: ServiceSchema = {
  name: 'www',

  settings: {
    port: 3000,
  },

  created() {
    this.server = http.createServer(this.httpHandler);
  },

  async started() {
    this.server.listen(this.settings.port);
  },

  async stopped() {
    this.server.close();
  },

  methods: {
    httpHandler(req: any, res: any) {
      res.end('Hello Moleculer!');
    },
  },
};
```

Here is the ts error:

```
Type '() => Promise<void>' is not assignable to type '() => Bluebird<void>'.
  Type 'Promise<void>' is not assignable to type 'Bluebird<void>'.
    Property 'caught' is missing in type 'Promise<void>'.
Type '() => Promise<void>' is not assignable to type '() => Bluebird<void>'.
  Type 'Promise<void>' is not assignable to type 'Bluebird<void>'.
```

Bluebird itself implements PromiseLike interface, so I thought It would be good choice to use PromiseLike instead of Bluebird to remove incompatibility. Or we could use union type, maybe?

### :gem: Type of change

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,3 @@
-import * as Bluebird from "bluebird";
 declare namespace Moleculer {
 	type GenericObject = { [name: string]: any };
 
@@ -22,7 +21,7 @@ declare namespace Moleculer {
 		trace(...args: any[]): void;
 	}
 
-	type ActionHandler<T = any> = ((ctx: Context) => Bluebird<T> | T) & ThisType<Service>;
+	type ActionHandler<T = any> = ((ctx: Context) => PromiseLike<T> | T) & ThisType<Service>;
 	type ActionParamSchema = { [key: string]: any };
 	type ActionParamTypes = "boolean" | "number" | "string" | "object" | "array" | ActionParamSchema;
 	type ActionParams = { [key: string]: ActionParamTypes };
@@ -96,7 +95,7 @@ declare namespace Moleculer {
 		cachedResult: boolean;
 
 		setParams(newParams: P, cloning?: boolean): void;
-		call<T = any, P extends GenericObject = GenericObject>(actionName: string, params?: P, opts?: GenericObject): Bluebird<T>;
+		call<T = any, P extends GenericObject = GenericObject>(actionName: string, params?: P, opts?: GenericObject): PromiseLike<T>;
 		emit(eventName: string, data: any, groups: Array<string>): void;
 		emit(eventName: string, data: any, groups: string): void;
 		emit(eventName: string, data: any): void;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -173,13 +172,13 @@ declare namespace Moleculer {
 		broker: ServiceBroker;
 		logger: LoggerInstance;
 		actions?: ServiceActions;
-		Promise: typeof Bluebird;
+		Promise: typeof Promise;
```
rmccallum81: I think it would be better to use `PromiseConstructorLike` here since typing it as `typeof Promise` (which is the same as `PromiseConstructor`) precludes it being a `Bluebird` style promise.

snnsnn: I changed `Promise` to `PromiseConstructorLike`, but it gives the error below. I had that error before and kept ` Promise` to satisfy the type guard:

```index.d.ts:175:19 - error TS2693: 'PromiseConstructorLike' only refers to a type, but is being used as a value here.
175  	Promise: typeof PromiseConstructorLike;
     	                ~~~~~~~~~~~~~~~~~~~~~~
index.d.ts:380:19 - error TS2693: 'PromiseConstructorLike' only refers to a type, but is being used as a value here.
380  	Promise: typeof PromiseConstructorLike;
     	                ~~~~~~~~~~~~~~~~~~~~~~
```
Do we need type guard for Promise here? What if we removed them?

```typescript
class ServiceBroker {
		constructor(options?: BrokerOptions);

		Promise: typeof PromiseConstructorLike;
		// ...
}
```

```typescript
class Service implements ServiceSchema {
		constructor(broker: ServiceBroker, schema?: ServiceSchema);

		protected parseServiceSchema(schema: ServiceSchema): void;

		name: string;
		version?: string | number;
		settings: ServiceSettingSchema;
		metadata: GenericObject;
		dependencies: string | GenericObject | Array<string> | Array<GenericObject>;
		schema: ServiceSchema;
		broker: ServiceBroker;
		logger: LoggerInstance;
		actions?: ServiceActions;
		Promise: typeof PromiseConstructorLike;

		waitForServices(serviceNames: string | Array<string> | Array<GenericObject>, timeout?: number, interval?: number): PromiseLike<void>;

		_init(): void;
		_start(): PromiseLike<void>;
		_stop(): PromiseLike<void>;
		[name: string]: any;
	}
```

icebob:
Thank you

Issue #436, opened 2018-12-14
Create an event between applyMixins and registerService


flaviocordova:
When a service using Mixins is created, the broker applies all the Mixins in the scheme, then registers the new service and only then call `created` method on that service.

The idea here is to create a new event so mixins could modify the scheme before it's registered. It could allow some interesting features like injecting actions or dependencies, or even wrapping actions so the mixins could do something before/after its invocation.

intech:
@flaviocordova please, see PR https://github.com/moleculerjs/moleculer/pull/444

icebob:
New middleware hook `serviceCreating` will be available in 0.14.

PR #444, opened 2019-01-08
Mixins hooks before and after by issue #436

intech:
### :dart: Relevant issues

#436 

### :gem: Type of change

### :scroll: Example code
```js
const mixin = {
    name: "mixin1",
    mixinBefore(parentSchema) { },
    mixinAfter(parentSchema) {}
};
``` 

- [x] Test Service mixins (should call every start handler) in https://github.com/moleculerjs/moleculer/blob/master/test/integration/service-mixins.spec.js

### DIFF
--- a/dev/dev.js
+++ b/dev/dev.js
@@ -4,6 +4,9 @@ const broker = new ServiceBroker({
 	//namespace: "test",
 	nodeID: "broker-1",
 	transporter: "NATS",
+	transit: {
+		disableVersionCheck: true
+	} ,
 	middlewares: [
 		{
 			localCall: (next) => {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,3 @@
-import * as Bluebird from "bluebird";
 declare namespace Moleculer {
 	type GenericObject = { [name: string]: any };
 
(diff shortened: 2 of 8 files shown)


icebob:
Thanks, interesting PR. But it is necessary to call the hooks at every merge? I mean, if you have a service with 
```js
{
	mixins: [
		mixin1,
		mixin2,
		mixin3
	]
}
```
It will be called 3 times. In my mind, I planned to add an after hook after all mixins are mixed.

intech:
> It will be called 3 times. In my mind, I planned to add an after hook after all mixins are mixed.

I can redo, but now the call goes through the chain using the example of the test:
1) mixin2L1 -> mixinL2
2) main -> mixinL2
3) mixin1L1 -> mixinL2
4) main -> mixinL2
in the argument is passed the parent scheme, which allows you to make more flexible hooks.
@icebob @flaviocordova what do you think?

icebob:
I will try it because I can't imagine where we can add the hooks. E.g. can we add these hooks into the main service file too, or just in the mixins? At present now I suppose that it can't solve a problem when you want to define a hook in the main class (probably without mixins) which is called before the service created from the schema.

PR #446, opened 2019-01-14
Fixed typescript type bug. Middleware type had legacy syntax.

jaybrto:
Typings for middleware was using old legacy syntax. Added two new types. Tested it using typescript build.


icebob:
Thanks, but it's not correct. The new middleware has several hooks but different params. Here is the example: https://github.com/moleculerjs/moleculer/blob/master/src/middleware.js#L146
Could you update to cover all hook?

jaybrto:
You got it. Using molecular i8 heavily now. Great job! I'll be contributing more add as I go. Wanted to choice something small like this in the mean time.

darky:
It obsolete, can be closed

PR #465, opened 2019-02-12
`ActionParamTypes` uses all types from "fastest-validator"

darky:


darky:
> Would be better to add all rule types from here: https://github.com/icebob/fastest-validator/blob/master/lib/validator.js#L9-L22

updated

Issue #489, opened 2019-03-11
Cache stampede


tiaod:
Consider the flowing service:
```js
module.exports = {
	name: "greeter",
	actions: {
		welcome: {
			params: {
				name: "string"
			},
			cache: {
				ttl: 5
			},
			handler(ctx) {
				console.log('run welcome')
				return new Promise(function(resolve, reject) {
					setTimeout(()=>resolve(`Welcome1, ${ctx.params.name}`), 3000)
				});
			}
		}
	}
};
```

## Expected Behavior
If cache expired, concurrent request will only call the `welcome` handler for only one time.
```text
[2019-03-11T10:19:30.566Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[2019-03-11T10:19:30.566Z] INFO  desktop-6avpisg-1828/API:    Call 'greeter.welcome' action
run welcome
[2019-03-11T10:19:30.701Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[2019-03-11T10:19:30.702Z] INFO  desktop-6avpisg-1828/API:    Call 'greeter.welcome' action
[2019-03-11T10:19:30.830Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[2019-03-11T10:19:30.831Z] INFO  desktop-6avpisg-1828/API:    Call 'greeter.welcome' action
[2019-03-11T10:19:30.957Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[2019-03-11T10:19:30.958Z] INFO  desktop-6avpisg-1828/API:    Call 'greeter.welcome' action
[2019-03-11T10:19:31.086Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[2019-03-11T10:19:31.087Z] INFO  desktop-6avpisg-1828/API:    Call 'greeter.welcome' action
[2019-03-11T10:19:31.220Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[2019-03-11T10:19:31.221Z] INFO  desktop-6avpisg-1828/API:    Call 'greeter.welcome' action
[2019-03-11T10:19:31.352Z] INFO  desktop-6avpisg-1828/API: => GET /api/greeter/welcome?name=John1
[...]

PR #490, opened 2019-03-11
Add cache lock

tiaod:
## :memo: Description

- [x] Add `lock` & `tryLock` method on cacher.
- [x] In cacher middleware, if a key of cache not found, will try to acquire a lock before calling the action handler.
- [x] Auto refresh stale cache by acquiring a lock.
- [x] Add lock & tryLock for the memory cacher
- [x] Add lock & tryLock for memory-lru cacher
- [x] Add lock & tryLock for redis cacher
- [x] Add `getWithTTL` api for cache, which act like `get` but also return the `ttl` of a key.

moleculer-cache-lock2

### :dart: Relevant issues

#489 
### :gem: Type of change

### :scroll: Example code
Cache config: 
```js
cacher: {
  ttl: 60,
  lock: true, // Set to true to enable cache locks. Default is disabled.
}

//  Or
cacher: {
  ttl: 60,
  lock: {
    ttl: 15, //the maximum amount of time you want the resource locked in seconds
    staleTime: 10, // If the ttl is less than this number, means that the resources are staled
  }
}

// Disable the lock
cacher: {
  ttl: 60,
  lock: {
    enable: false, // Set to false to disable.
    ttl: 15, //the maximum amount of time you want the resource locked in seconds
    staleTime: 10, // If the ttl is less than this number, means that the resources are staled
  }
}
```

For redis cacher, add `redlock` options:
```js
  cacher: {
    type: "Redis",
    options: {
      // Prefix for keys
      prefix: "MOL",
      // set Time-to-live to 30sec.
      ttl: 30,
      // Turns Redis client monitoring on.
      monitor: false,
      // Redis settings
      redis: {
        host: "redis-server",
        port: 6379,
        password: "1234",
        db: 0
      },
      lock: {
        ttl: 15, //the maximum amount of time you want the resource locked in seconds
        staleTime: 10, // If the ttl is less than this number, means that the resources are staled
      },
      // Redlock settings
      redlock: {
        // Redis clients. Support node-redis or ioredis. By default will use the local client.
        clients: [client1, client2, client3],
        // the expected clock drift; for more details
        // see http://redis.io/topics/distlock
        driftFactor: 0.01, // time in ms

        // the max number of times Redlock will attempt
        // to lock a resource before erroring
        retryCount: 10,

        // the time in ms between attempts
        retryDelay: 200, // time in ms

        // the max time in ms randomly added to retries
        // to improve performance under high contention
        // see https://www.awsarchitectureblog.com/2015/03/backoff.html
        retryJitter: 200 // time in ms
      }
    }
  }
});
``` 
Example code to write a custom cacher:
```js
const BaseCacher = require("moleculer").Cachers.Base;

class MyCacher extends BaseCacher {
    get(key) { /*...*/ }
    set(key, data, ttl) { /*...*/ }
    del(key) { /*...*/ }
    clean(match = "**") { /*...*/ }
    // The following is only needed when you enable the cache lock.
    lock(key) { /*...*/ }
    tryLock(key) { /*...*/ }
    getWithTTL(key) { /*...*/ }
}
```
The `lock` method should return an `unlock` function which return a promise, then call `unlock()` to realse the lock.

The `tryLock` should throw a error and not blocking the request.

```js
// This will blocking the request, until get the lock.
let unlock = await cacher.lock(key)
// Do something here
await unlock()
```

```js
let unlock;
try{
  unlock = await cacher.tryLock('abc123')
  // You get the lock, do something here.

  // Finally
  await unlock()
}catch(e){
  // The lock is occupied. Continues...
}
```

The `getWithTTL` method is act like `get` but also return the `ttl` of a key.
```js
let data = await cacher.get('key')
let { data, ttl } = await cacher.getWithTTL('key')
```

Test middleware with lock enabled
- [x] It should give back the cached data and not called the handler
- [x] It should not give back cached data and should call the handler and call the 'cache.set' action with promise
- [x] It should disable cache lock by defalut
- [x] It should not call cacher.getWithTTL if cache.lock = false
- [x] It should not call cacher.getWithTTL if cache.lock = { enabled: false }
- [x] It should call the handler only once when concurrency call a cacher with lock
- [x] It should realse the lock when an error throw
- [x] It should refresh a stale key of cache
- [x] It should not call the handler if the cache is refreshed
- [x] It should realse the lock when refreshing a key and an error throw

Test lock method of `MemoryCacher` & `MemoryLRUCacher`
- [x] It should lock the concurrency call
- [x] It should reject the promise when failed to unlock

Test tryLock method of `MemoryCacher` & `MemoryLRUCacher`
- [x] should throw an error when already locked

Test `MemoryCacher` & `MemoryLRUCacher` getWithTTL method
- [x] It should return data and ttl

Test `RedisCacher` getWithTTL method
- [x] It should call the get and ttl using pipeline
- [x] It should throw an error when get method return error
- [x] It should throw an error when ttl method return error
- [x] It should return null when JSON.pares failed

Test `RedisCacher` lock method
- [x] It should call redlock.lock when calling cacher.lock
- [x] It should call redlock.unlock when calling unlock callback
- [x] It should call redlock.lock when calling cacher.tryLock
- [x] It should call redlock.unlock when calling unlock callback
- [x] It should failed to acquire a lock when redlock client throw an error

Test `RedisCacher` with opts.lock
- [x] It should create redlock clients when opts.lock==true
- [x] It should not create redlock clients when opts.lock==false
- [x] It should not create redlock clients when opts.lock.enabled==false


zllovesuki:
Wouldn’t it make more sense if you acquire the lock *after* the handler is called, the lock-save-unlock process is done asynchronously and independent of the handler call?

tiaod:
@zllovesuki How? Actually I'm not very good at dealing with locks. Could you show me some example code? 😺

zllovesuki:
It's more a design decision rather than actual code.

With the current approach (and Bulkhead enabled), all the requests following the first request will be blocked as they must wait for the first request to resolve, save, then release the lock. It may or may not be desirable depending on application needs. For example, this could be desirable if the action involves a complicated SQL query and takes a long time to complete. However, for the lesser resource intensive actions, it adds unnecessarily latency.

Although I suppose if the action calls a heavy query, developer/operator should be doing cache warming instead of relying on locking.

tiaod:
So I can add a flag to control the lock and disable it by default

tiaod:
After some search, I found out that if I want to acquire a lock after the handler is called, I need a way to get the ttl of the cached key, and the problem I'm facing called **dog-pile effect**
https://www.sobstel.org/blog/preventing-dogpile-effect/

tiaod:
Done. It's ready now

package.json
```diff
@@ -90,6 +91,7 @@
     "fastest-validator": "^0.6.15",
     "glob": "^7.1.3",
     "ipaddr.js": "^1.9.0",
+    "lock": "^1.1.0",
```
icebob: Please try to avoid to use a new dependency. I've checked the `lock` code and very simple. It stores the keys in an `Object`. However, it is not a Promise-based lib, so it generates a lot of code to convert it to promises. Would be good to create an own simple lock class, which store keys in a `Map` and all methods are Promise based. Could you make it?

src/cachers/redis.js
```diff
@@ -39,20 +39,36 @@ class RedisCacher extends BaseCacher {
...
 		this.client = new Redis(this.opts.redis);
+		let redlockClients = (this.opts.redlock ? this.opts.redlock.clients : null) || [this.client]
+		this.redlock = new Redlock(
+			redlockClients,
+			_.omit(this.opts.redlock, ['clients'])
+		);
+		this.redlockNonBlocking = new Redlock(
+			redlockClients,
+			{
+				retryCount: 0
```
icebob: This setting would be good for everyone (I don't know)? If not, would be good to expose to `this.opts.redlock.retryCount`

package.json
tiaod: OK, I will remove this lib.

src/cachers/redis.js
tiaod: Yes, this setting would be good for everyone. There are two kinds of redlock clients internal. 
1. `redlock` : will take the `this.opts.redlock` as options, used for `cacher.lock(key)` method.
2. `redlockNonBlocking` will take `{ retryCount: 0 }` as options, used for `cacher.tryLock(key)` method, which will not block the request.

tiaod:
Done, I had already remove the `lock` lib.

src/lock.js
```diff
@@ -0,0 +1,36 @@
+class Lock {
+  constructor(){
+    this.locked = new Map();
+  }
+
+  acquire(key, ttl){
+    let locked = this.locked.get(key);
```
icebob: `ttl` param is not used. Is it by design?

test/unit/lock.spec.js
```diff
@@ -0,0 +1,60 @@
+const Lock = require("../../src/lock");
+
+describe("Test lock", () => {
+  let key = "abc123"
+  it("should lock", () => {
+    const lock = new Lock();
+
+    return lock.acquire(key).then(() => {
+      expect(lock.isLocked(key)).toBeTruthy()
+      lock.release(key).then(() => {
```
icebob: missing `return` from the start of the line. It's important because if an error occurs, jest won't detect it and it will say the test is passed.

src/lock.js
```diff
@@ -0,0 +1,36 @@
...
+        locked.shift()(); // Release the lock
+      }else{
+        this.locked.delete(key);
+      }
+    }
+    return Promise.resolve();
+  }
+}
+
+module.exports = Lock;
```
icebob: Nice & simple implementation. Thanks

src/lock.js
tiaod: For a single process, ttl is not necessary. So I didn't add the ttl mechanism. I will comment it.

test/unit/lock.spec.js
tiaod: OK, I will fix it.

src/lock.js
icebob: OK

icebob:
Nice PR @tiaod, thank you!

PR #496, opened 2019-03-19
added seq property on PacketInfo protobuf message

alvarocantador:
## :memo: Description

I was trying to add new services dynamically after the broker has been started with `broker.createService` and the other nodes connected in the same transporter (NATS) doesn't update the services list. This bug happens only with the ProtoBuf serializer.

### :gem: Type of change

Just added the `seq` property to PacketInfo message, in order to turn possible the nodes updates their registered services.

This bug occurs since the gossip was updated from when to seq: https://github.com/moleculerjs/moleculer/commit/fbb99106150d4b41798b6d9550c547f83e9f2aa6

This is not a breaking change.

### :scroll: Example code
```js

``` 

- [ ] Test A
- [ ] Test B

### DIFF
--- a/src/serializers/proto/packets.proto
+++ b/src/serializers/proto/packets.proto
@@ -51,8 +51,7 @@ message PacketInfo {
 	repeated string ipList			= 5;
 	required string hostname		= 6;
 	required Client client 			= 7;
-	required int32  seq 			= 8;
-
+	optional int32  seq 			= 8;
 
 	message Client {
 		required string type 			= 1;
--- a/src/serializers/proto/packets.proto
+++ b/src/serializers/proto/packets.proto
@@ -51,6 +51,8 @@ message PacketInfo {
 	repeated string ipList			= 5;
 	required string hostname		= 6;
 	required Client client 			= 7;
+	required int32  seq 			= 8;
+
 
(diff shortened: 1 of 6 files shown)


icebob:
Nice catch. Interesting how I forgot it.

edurdias:
Does the same happen to Thrift transporter?

icebob:
Yes, it affects all schema-based serializers, ProtoBuf, Thrift & Avro. The bigger problem is Avro can't handle new properties, so it causes a breaking change so I can fix it in 0.14 only.

@alvarocantador could you check that it doesn't cause a problem that fixed & non-fixed nodes can communicate with each other?

edurdias:
@icebob it may be a good idea to release ProtoBuf and Thrift as a patch (for example, 0.13.8) and then do a breaking change on Avro. That would unblock developers using these 2 serializers.

alvarocantador:
@icebob Yes, we run both and they communicate with each other

icebob:
@edurdias yes, I will do this.
@alvarocantador thank you!

icebob:
@alvarocantador it doesn't work as `required`:
image

edurdias:
👍

Issue #453, opened 2019-01-23
Can't use Logger in Typescript


darky:
## Expected Behavior

`Logger` exported as typescript `interface`, but should be runtime value

## Current Behavior

```typescript
import { BrokerOptions, Logger } from 'moleculer';

const brokerOptions: BrokerOptions = {
  logger: (bindings) => Logger.extend(/* ... */)
  // [ts] 'Logger' only refers to a type, but is being used as a value here. [2693]
}
```

## Failure Information

`[ts] 'Logger' only refers to a type, but is being used as a value here. [2693]`

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.13.5

AndreMaz:
@darky  https://github.com/moleculerjs/moleculer/pull/452 solves this issue, right? If so, then this issue can be closed.

darky:
Not, it was other problem

Issue #434, opened 2018-12-11
Typescript for winston Logger


huycuong89:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I checked the documentation and found no answer

## Expected Behavior

When add "winston" as logger in broker's options, it should be allowed.

## Current Behavior
"winston" not typed as "Logger", Typescript throw error.

## Failure Information

The expected type comes from property 'logger' which is declared here on type 'BrokerOptions'

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Create "winston" logger
2. Add "winston" to broker's options

### Reproduce code snippet
```js
const loggerWinter: WTLogger = winston.createLogger({
  level: 'info',
  format: winston.format.combine(winston.format.timestamp(), winston.format.json()),
  transports: [
    new winston.transports.File({ filename: 'error.log', level: 'error' }),
    new winston.transports.File({ filename: 'combined.log' }),
  ],
});

const brokerConfig: BrokerOptions = {
  namespace: '',
  nodeID: (NODEID ? NODEID + '-' : '') + os.hostname().toLowerCase(),

  logger: loggerWinter, // WARNING FROM TYPESCRIPT
  logLevel: 'info',
  logFormatter: 'default',
  // other options
}
```

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.13
* NodeJS version: 10.13.0
* Operating System: Ubuntu 18.04 x64

darky:
tablet for this here: https://github.com/moleculerjs/moleculer/pull/501/

darky:
@icebob, it can be closed.
Will be fixed in version `> 0.13.8`

darky:
Example: https://moleculer.services/docs/0.13/logging.html#Winston

PR #501, opened 2019-03-24
improve logger types

darky:
## :memo: Description

Now you can use Logger in TypeScript properly.
Also add tsd for typing tests

### :dart: Relevant issues

#453 
#434 

### :gem: Type of change

Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test:int": "jest --testMatch \"**/integration/**spec.js\" --coverage --no-cache",
     "test:amqp": "jest --testMatch \"**/transporters/amqp/**spc.js\" --runInBand",
     "test:trans": "jest --testMatch \"**/transporters/index.spc.js\"",
-    "test:ts": "tsc -p test/typescript/hello-world && ts-node -P test/typescript/hello-world/tsconfig.json test/typescript/hello-world/index.ts",
+    "test:ts": "tsd && tsc -p test/typescript/hello-world && ts-node -P test/typescript/hello-world/tsconfig.json test/typescript/hello-world/index.ts",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -98,7 +101,10 @@
     "node": ">= 6.x.x"
   },
   "tonicExampleFilename": "docs/runkit/simple.js",
-  "typings": "./index.d.ts",
+  "types": "./index.d.ts",
```
icebob: Is it a common change from `typings` to `types`?

package.json
```diff
@@ -77,6 +79,7 @@
     "redlock": "^3.1.2",
     "thrift": "0.12.0",
     "ts-node": "8.0.2",
+    "tsd": "0.7.0",
```
icebob: `tsd` is a [deprecated library](https://github.com/DefinitelyTyped/tsd).

darky: No, seems you mean this:
https://github.com/DefinitelyTyped/tsd

But used this:
https://github.com/SamVerschueren/tsd

I had seen, that @sindresorhus use it in https://github.com/sindresorhus/type-fest and think, that it good library for testing typescript declaration. Anyway, it better, that before with `hello-world` :)

package.json
darky: `tsd` say me to use "types" instead "typings"
Maybe it something related to this:
https://stackoverflow.com/questions/39261204/typings-vs-types-npm-scope

package.json
icebob: Ohh, sorry. Yes, I found the other.

package.json
icebob: Thanks.

Issue #483, opened 2019-02-25
Service modifications support for ES6 classes


akazakou:
**Is your feature request related to a problem? Please describe.**
I want to isolate the configuration mechanism from service initialization. In that case, I want to change `settings` property of schema. Regarding to [ES6 classes documentation](https://moleculer.services/docs/0.13/services.html#ES6-classes) I can't use schema modifiers. But `brocker.createService` supporting the second parameter with service modifications. 

I want to implement the second parameter of `Service` class constructor to support schema modifications for ES6 services.

**Describe the solution you'd like**
Update `brocker.createService` functionality to pass schema modifications into creating a new instance of service as the second parameter of `Service` class constructor
```typescript
constructor(broker: ServiceBroker, schemaMods?: ServiceSchema) {}
```

**Describe alternatives you've considered**
Right now I'm using a static property of Service to store schema modifications and apply it in a constructor:
```typescript
class ServiceSession extends Service {
  public static schemaMods: ServiceSchema = {} as ServiceSchema;

  constructor(broker: ServiceBroker) {
    super(broker);

    const serviceSchema: ServiceSchema = {
      /**
       * Name of current service
       */
      name: packageJson.name,
      // ***
    };

    // compile service schema and redefine settings
    this.parseServiceSchema(Object.assign({}, serviceSchema, ServiceSession.schemaMods));
  }
}
```

**Additional context**
The main idea of this feature request isolates Service functionality from environment configuring. Give DevOps team possibility to configure and modify service behavior, without requesting of Development team to code changes. 

It should give the Development team a possibility to isolate business logic inside the service. All API configuration, settings providers, additional middleware and etc should be configured outside of the service.

akazakou:
I guess, it also should be applied to `brocker.loadService` and `brocker.watchService` function too.
`brocker.loadServices` should support object configuration with a key of the service name.

Maybe some implementations too, if it related to service initialization.

fugufish:
Maybe, I don't understand the suggestion, but couldn't settings be modified per-environment just as easily using environment variables by devops without developer intervention?

akazakou:
We have many ways how we can configure our services. Example: AWS Parameters Store, Some specific configuration service, environment variables or other. In that case, if we have different configuration flows for different environments like localhost, test, staging, the production we should support many settings sources in service code.

I want to move all configuration stuff to service runner. And implement separate service runner for each environment. Example:
* localhost environment will connect all available services and configure it from service runner configuration file
[...]

PR #514, opened 2019-04-05
Service modifications support for ES6 classes

akazakou:
## :memo: Description

In some cases architectual solutions can split service and service runner logic. In that case, if you need provide to service some settings or update schema parameters, you can use `schemaMods`. 

It working for GenericObject definition, but not working for ES6 classes. I added providing `schemaMods` parameters to ES6 classes.

### :dart: Relevant issues
#483 Service modifications support for ES6 classes

### :gem: Type of change

### :scroll: Example code
```js
class ES6Service extends Service {
	constructor(broker, schemaMods) {
		super(broker);

		this.name = "es6-with-schema-mods";
		this.version = 2;
		this.settings = {};
		this.actions = {
			send: () => Prmise.resolve()
		};

		if (schemaMods && schemaMods.settings && schemaMods.settings.connectionString) {
			this.settings.connectionString = schemaMods.settings.connectionString;
		}

		// this should prevent override this.actions in parseServiceSchema method
		this.parseServiceSchema(Object.assign({}, this));
	}
}

let broker = new ServiceBroker({ logger: false });
broker.createService(ES6Service, { settings: { connectionString: "mongodb://localhost:27017/test" } });
``` 

I've added 2 additional unit tests to check schema creating proccess for ES6 classes

- [x] should create service from ES6 instance without schema mods
- [x] should create service from ES6 instance with schema mods

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -39,11 +39,11 @@ class Service {
 	 * Creates an instance of Service by schema.
 	 *
 	 * @param {ServiceBroker} 	broker	broker of service
-	 * @param {Object=} 		schema	schema of service
+	 * @param {Object} 			schema	schema of service
 	 *
 	 * @memberof Service
 	 */
-	constructor(broker, schema = undefined) {
+	constructor(broker, schema) {
 		if (!_.isObject(broker))
 			throw new ServiceSchemaError("Must set a ServiceBroker instance!");
 
--- a/src/service.js
+++ b/src/service.js
@@ -39,11 +39,11 @@ class Service {
 	 * Creates an instance of Service by schema.
 	 *
 	 * @param {ServiceBroker} 	broker	broker of service
-	 * @param {Object} 			schema	schema of service
+	 * @param {Object=} 		schema	schema of service
 	 *
 	 * @memberof Service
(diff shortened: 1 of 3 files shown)


src/service.js
```diff
@@ -39,11 +39,11 @@ class Service {
 	 * Creates an instance of Service by schema.
 	 *
 	 * @param {ServiceBroker} 	broker	broker of service
-	 * @param {Object} 			schema	schema of service
+	 * @param {Object=} 		schema	schema of service
 	 *
 	 * @memberof Service
 	 */
-	constructor(broker, schema) {
+	constructor(broker, schema = undefined) {
```
icebob: Why need it `schema = undefined`. I think it is not relevant, because if you don't define˙it, it will be `undefined`

akazakou: My IDE highlit this. I just added this for comfortable IDE workaround. If it not required, I can remove this

icebob: OK.

akazakou: @icebob I reverted changes in `service.js` file. Right now only related to feature changes

PR #524, opened 2019-04-20
Fix Cachers TS definitions

shawnmcknight:
## :memo: Description

The typescript definitions for `Cachers` is incorrect.  It has no reference for the `Base` cacher and the references for the `Memory` and `Redis` cachers are indicated as instances of `Cacher` rather than a class constructor.  This prevents creating a custom cacher in TypeScript since you cannot extend what isn't there.

This PR updates the definitions so that `Base`, `Memory`, and `Redis` are all class constructors rather than instances.

### :gem: Type of change

This really _shouldn't_ be a breaking change, but its possible that a consumer was using the existing `Cacher` or `Cachers` types in some strange way.

Typescript tests included in PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -627,16 +627,24 @@ declare namespace Moleculer {
 		deserialize(type: string, data: Buffer): Packet;
 	}
 
-	class Cacher {
-		constructor(opts?: GenericObject);
-		init(broker: ServiceBroker): void;
-		close(): PromiseLike<any>;
-		get(key: string): PromiseLike<null | GenericObject>;
-		set(key: string, data: any, ttl?: number): PromiseLike<any>;
-		del(key: string|Array<string>): PromiseLike<any>;
-		clean(match?: string|Array<string>): PromiseLike<any>;
-		client?: any;
+	namespace Cachers {
+		class Base {
+			constructor(opts?: GenericObject);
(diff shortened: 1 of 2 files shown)


icebob:
Thanks, so it means, I can merge it only in 0.14?

shawnmcknight:
> Thanks, so it means, I can merge it only in 0.14?

Technically, the type is totally different now.  With that said, I'm honestly not sure how anyone could have been using it the way it was previously.  So, technically it would be a breaking change for anyone who managed to use it, but they were probably doing something weird with it.

Safest bet is to make it 0.14 rather than 0.13.x, but you could probably get away with it being in 0.13.x.

shawnmcknight:
> Safest bet is to make it 0.14 rather than 0.13.x, but you could probably get away with it being in 0.13.x.

It might be a moot point anyway.  I'll be submitting another PR for types after this one merges in which would _definitely_ be a breaking change if accepted.  We'll have to decide whether having more appropriate type names is worth the breaking change or not.  I think it would be, but it'll make the decision easier on what version to do here if they gets in also.

icebob:
In this case, please send this PR to the `next` branch.

shawnmcknight:
Closing in favor of #525 which is targeting `next` branch.

PR #525, opened 2019-04-22
Fix Cachers TS definitions

shawnmcknight:
## :memo: Description

The typescript definitions for `Cachers` is incorrect.  It has no reference for the `Base` cacher and the references for the `Memory` and `Redis` cachers are indicated as instances of `Cacher` rather than a class constructor.  This prevents creating a custom cacher in TypeScript since you cannot extend what isn't there.

This PR updates the definitions so that `Base`, `Memory`, and `Redis` are all class constructors rather than instances.

### :gem: Type of change

Typescript tests included in PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -626,16 +626,24 @@ declare namespace Moleculer {
 		deserialize(type: string, data: Buffer): Packet;
 	}
 
-	class Cacher {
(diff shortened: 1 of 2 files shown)


icebob:
Thank you!

PR #582, opened 2019-08-06
Watch services folder

igk1972:
The changes are made to allow you to add service files to a broker that is already running.

Before that, the list of service files is processed only at the start of the broker (or moleculer-runner).

### :gem: Type of change

### :scroll: Example code
```sh
  npx moleculer-runner -hot /foo/bar/services/ /others/some/services/
``` 

- [ ] Test A.
1. Create folder tmp/test
2. Start ```'moleculer-runner -hot tmp-test/```.
3. In other terminal, copy ready service(s) to folder tmp-test.
4. Expect: service(s) loaded and started.
5. Delete service in folder tmp-test.
6. Expect: service stoppedn and unloaded without error.


icebob:
Thanks @igk1972!
@AndreMaz will review & merge it.

AndreMaz:
Hey @igk1972 . Sorry for the delay.

I've tested but something isn't right because I've got an error.
Steps to reproduce:
1. Create [moleculer-demo](https://moleculer.services/docs/0.13/usage.html#Create-a-Moleculer-project) project
2. run `npm run dev`
3. create new service as follows:
3UTUjVWiy0

AndreMaz:
I'm also not a huge fan of the place where the code is being executed for 2 reasons:
1) It's being registered as a custom middleware despite (if merged) being part of the moleculer's core
image
2) The file/folder watching and loading logic is scattered across multiple files [`hot-reload.js`](https://github.com/moleculerjs/moleculer/blob/next/src/middlewares/hot-reload.js) and `moleculer-runner.js `. I would prefer having everything at a single place -> `hot-reload.js`

igk1972:
> I've tested but something isn't right because I've got an error.

Hi @AndreMaz, sorry for the delay.

The changes assume that the service folder creates a ready-made service file.
Case development - the service files will be loaded after each saving.
Case production - the service folder will be populated by service files that have already been tested separately in other locations. Maybe someone will need this feature.

This is a useful feature - not to restart the moleculer-runner.

Personally, I made these changes to be able to compile service files using webpack. That's why you need to do something with moleculer-runner. Since you need to be able to create (or transfer) watchers for services.

Middleware hot-reload.js only works for a specific service. And does not know anything about service folders.

icebob:
I agree with @AndreMaz, would be better to move this logic into Hot Reload middleware. We should transfer the watching folders from runner to the Hot-Reload middleware. It could be (e.g.) a custom broker options property.

AndreMaz:
> This is a useful feature - not to restart the moleculer-runner.

I absolutely agree with you. This feature is very useful.

> Case production - the service folder will be populated by service files that have already been tested separately in other locations. 

In my test I've used the following service
```js
"use strict";

module.exports = {
    name: "test"
};
```
It's a valid service schema so it should load the service after I've saved it in `services` folder. However,   instead of loading the service it produced an error... That's why I've said that something isn't 100% right.

> Middleware hot-reload.js only works for a specific service. And does not know anything about service folders.

You are right again but your solution would increase technical debt. Moving the watchers to `hot-reload.js` will require more work but it would be a "cleaner" solution

icebob:
FYI: I'm working on to move the logic to hot-reload middleware...

icebob:
It's done. The extra feature is that the hot-reload watches moleculer.config.js file, as well. If it changed, it restarts the broker.
Another new feature is that when loading a newly created services & loading is failed, it will try to load again if changed. It can be good, when you create a new file in your IDE it will be empty, but HotReload tries to load but it'll fail, so you write the code of services, save, and HotReload tries to load again. In previous versions, it didn't work. You should have restarted the Runner always.

AndreMaz:
Awesome work icebob 😄

igk1972:
Sorry for delay.

Thanks guys for quick smart solution.

PR #588, opened 2019-08-21
Fix protobuf definitions

fugufish:
## :memo: Fixes incorrect proto 3 definition

The Protobuf file is listed as using proto3, but the syntax is incorrect. It seems that protobufjs doesn't care, but protoc does. We are working on implementing the protobuf serializer for moleculer-ruby which requires protoc, and we want full parity between the packets. To do this we are actually curling the proto package from moleculerjs and using that, but the definition file is syntactically incorrect for proto3.

My real suggestion would be to create a new protocol repo, that includes the protocol document itself and the packet definition files for serializers like protobuf.

This should not be a breaking change, as it simply removes the required and optional flags (not supported in proto3).


icebob:
Hmm, good to know, thanks!

PR #589, opened 2019-08-24
Use serializers in redis

shawnmcknight:
## :memo: Description

The existing `Redis` cacher _always_ serializes using JSON.  However, there are differences between how JSON serialization works and how some of the other transporter serializers work.  For instance, the Notepack and Msgpack serializers are able to serialize and deserialize `Date` objects while JSON serialization converts them to date-strings.

Because the `Redis` cacher serializes using JSON regardless of the transporter's serializer, this can result in different behaviors when an action result pulls from cache instead of being run through the action normally.  For instance, if an action returns an object with a `Date` object for one of its properties, when using Notepack this will be returned as a true `Date` object.  However, if the same action responds using the Redis cache, it will have been converted to a string.

This PR adds handling to the `Redis` cacher to accept an option named `serializer`.  This `serializer` option works in exactly the same manner as the transporter serializer does.  By specifying an option for this value, that serializer will be used when serializing/deserializing to and from the Redis cache.  It would be recommended to keep the transporter and cacher serializer in sync to avoid any strange behaviors.

If a cacher serializer is not chosen, JSON will default (just like the transporter), so this does not cause any breaking change.

While making these changes I found a large number of errors reported by `eslint` which were also addressed.

### :gem: Type of change

### :scroll: Example code
```js
  cacher: {
    type: "Redis",
    options: {
      serializer: "Notepack",
    },
  },
});
``` 

Redis cacher unit tests have been updated as part of the PR.

I also ran a modified version of the `simple` example service to turn on caching and run the same action call repeatedly.  I ran tests without specifying a serializer (which should use JSON) as well as tests with specifying the Notepack serializer.  I confirmed that the action results from the first call were the same as the repeated calls which would have come from cache.  I also used the Redis CLI to confirm that the data was being stored differently when using the different serializers.

### DIFF
--- a/dev/redis-cacher.js
+++ b/dev/redis-cacher.js
@@ -1,0 +1,46 @@
+"use strict";
+
+let chalk = require("chalk");
+let ServiceBroker = require("../src/service-broker");
+
+// Create broker
+let broker = new ServiceBroker({
+	logLevel: "debug",
+	cacher: {
+		type: "Redis",
+		options: {
+			serializer: "Notepack"
+		}
+	}
+});
+
+broker.createService({
+	name: "greeter",
+	actions: {
+		hello: {
+			//cache: false,
+			cache: {
+				enabled: true
+			},
+			handler(ctx) {
+				return {
+					a: 5,
+					b: "John",
+					c: true,
+					d: null,
+					e: {
+						f: Date.now(),
+						g: new Date()
+					},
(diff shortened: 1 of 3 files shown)


icebob:
Good job, I'm checking...

icebob:
[Released in 0.13.10](https://github.com/moleculerjs/moleculer/releases/tag/v0.13.10)

PR #615, opened 2019-11-11
Add stop method to metrics reporters

AndreMaz:
## :memo: Description

Adding `stop()` method to metrics reporters. 

Stop method is not particularly important for any Push-based reporters as there is nothing to "stop".  However, for Prometheus, which uses pull mechanism, `stop()` is necessary to gracefully shutdown the http server.

This PR fixes the error that occurred when `moleculer.config.js` was (hot)reloaded with Prometheus metrics enabled.

### :gem: Type of change


test/unit/metrics/registry.spec.js
```diff
@@ -111,6 +111,21 @@ describe("Test Metric Registry", () => {
...
+			const broker = new ServiceBroker({ logger: false });
+			const metric = new MetricRegistry(broker, {
+				collectProcessMetrics: true,
+				reporter: "Event"
+			});
+
+			metric.init();
+
+			expect(metric.stop()).toBeInstanceOf(broker.Promise);
+		});
```
icebob: You have to check that the reporter `stop` method is called.
`metric.reporters[0].stop = jest.fn(() => Promise.resolve())`
And after stop executed
`expect(metric.reporters[0].stop).toHaveBeenCalledTimes(1);`

AndreMaz: Nice catch. I've added

PR #624, opened 2019-11-27
Updates to 0.14 typedefs of Context-based events and debounce and thr…

artur-krueger:
## :memo: Description
Changed typedef of ServiceEventHandler to match old and new signature.
Added 0.14 options (context, debounce, throttle) of ServiceEvent.

This change is neccessary to be able to use the new options and Conext-based events with TypeScript.

### :gem: Type of change

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -534,10 +534,10 @@ declare namespace Moleculer {
 		context?: boolean;
 		debounce?: number;
 		throttle?: number;
-		handler?: ServiceEventHandler | ServiceEventOldHandler;
+		handler?: ServiceEventHandler | ServiceEventLegacyHandler;
 	}
 
-	type ServiceEvents = { [key: string]: ServiceEventHandler | ServiceEventOldHandler | ServiceEvent };
+	type ServiceEvents = { [key: string]: ServiceEventHandler | ServiceEventLegacyHandler | ServiceEvent };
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -524,15 +524,20 @@ declare namespace Moleculer {
 		[name: string]: any;
 	}
 
-	type ServiceEventHandler = ((payload: any, sender: string, eventName: string) => void) & ThisType<Service>;
+	type ServiceEventOldHandler = ((payload: any, sender: string, eventName: string, ctx: Context) => void) & ThisType<Service>;
```
shawnmcknight: In keeping with how the below 0.14 signature is documented as "Legacy", I would name this type `ServiceEventLegacyHandler` to avoid any potential confusion.

index.d.ts
```diff
@@ -524,15 +524,20 @@ declare namespace Moleculer {
...
+
+	type ServiceEventHandler = ((ctx: Context) => void) & ThisType<Service>;
```
shawnmcknight: It should be noted in the PR that this is a _breaking_ change.  Since 0.14 is currently in beta, that should be fine and it is a good time to get the breaking change in place.

artur-krueger: Renamed as suggested and added note about being a breaking change. Thank you

PR #626, opened 2019-11-29
Update types for Context, ActionHandler, and Cachers Base

glenn-p-ryan:
## :memo: Description
This PR updates types for `Context` and `ActionHandler` which, as a result, should solve the problem discussed in [PR 600](https://github.com/moleculerjs/moleculer/pull/600). Additionally, the type `defaultKeygen` has been added to the `Cachers` namespace under the `Base` class. This method type for generating the key seems to be missing.

### :dart: Relevant issues
Issue being discussed [here](https://github.com/moleculerjs/moleculer/pull/600)

### :gem: Type of change

Overrode types in our current implementation of moleculer and tested for ts errors.


icebob:
@ipetrovic11 @shawnmcknight please review it.

icebob:
I've checked the source, so in the `defaultKeygen` method, the `params` can be null. But if it's not null, it should be an `object`.So, the d.ts modification is correct, I should update the jsdoc according to it.

glenn-p-ryan:
In the `defaultKeygen` method, does the first condition:
`if(params || meta){...}`
also warrant the need to update the jsdocs for `meta` to possibly be null as well?

shawnmcknight:
> In the `defaultKeygen` method, does the first condition:
> `if(params || meta){...}`
> also warrant the need to update the jsdocs for `meta` to possibly be null as well?

It appears to me that `meta` will default to empty object in the `Context` constructor and will only be mutated if the parent context's `meta` is not loose null or the calling options are not loose null.  That implies to me that there is no way for `meta` is ever be `null` by the time it gets to a called action, so that check seems unneeded and there shouldn't be a need to document `meta` in that fashion.

@icebob Do you agree?

icebob:
Yes, meta can't be null, it's initialized as `{}`. Thanks @glenn-p-ryan

PR #656, opened 2020-01-23
update fastest-validator 1.0.0

intech:
All tests passed.
Add tag v0.13.13

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3813,9 +3813,9 @@
       "dev": true
     },
     "fastest-validator": {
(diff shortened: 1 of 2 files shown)


icebob:
We can't add the fastest-validator 1.0.0 to the 0.13 version because it contains breaking changes (manipulate the original object) which causes breaking changes in Moleculer as well.. But in 0.14 it will be the default version.

If you want to use 1.0.0 in 0.13 you can do it this way: https://repl.it/@icebob/moleculer-validator-100

PR #681, opened 2020-02-24
Allow custom logging for "broker.call"

dordai-robert:
… Initial proposition

## :memo: Description

Hi @icebob we're using moleculer for a project in production, and our kibana logs are starting to get pretty overwhelming. We have a lot of "Call action locally" logs and being limited to only "action" and "requestID" when logging our action calls is making debugging a lot harder(Yes, we can make some pre hooks to log some more info but I still think having the option to customize "Call action locally" call would be the cleanest way). Having this option to send whatever object we want for the "Call action locally" logs would make our lifes a lot easier.

This is not a ready for merge PR, I have not spent enough time in the code to see everything that needs checking for a change like this. I wanted to post this initial PR to get your thoughts on this, to know if it's worth to invest more time into this. Or if you have any other solution that gives us the option I mentioned would be great.

### :gem: Type of change

### :scroll: Example code
```js

``` 

- [ ] Test A
- [ ] Test B

### DIFF
--- a/src/service-broker.js
+++ b/src/service-broker.js
@@ -1048,9 +1048,9 @@ class ServiceBroker {
 		}
 
 		if (ctx.endpoint.local)
-			this.logger.debug("Call action locally.", { action: ctx.action.name, requestID: ctx.requestID });
+			this.logger.debug("Call action locally.", opts.customLocalLoggingCallObj || { action: ctx.action.name, requestID: ctx.requestID });
 		else
-			this.logger.debug("Call action on remote node.", { action: ctx.action.name, nodeID: ctx.nodeID, requestID: ctx.requestID });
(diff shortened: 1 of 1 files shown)


icebob:
Hi,
I don't think so that we should add this special thing to the core of Moleculer because you can do similar things with middleware. You can wrap the action handlers or the `broker.call`. Here is an example for both:

```js
    middlewares: [{
        localAction(handler, action) {
            return ctx => {
                this.logger.info(`> Action '${action.name} 'called with`, ctx.params);
                return handler(ctx);
            };
        },

        call(next) {
            return (actionName, params, opts) => {
                this.logger.info(`>> Action '${actionName} 'called with`, params);
                return next(actionName, params, opts);
            };
        }
    }]
```

dordai-robert:
I understand. I tried actually using middlewares but that just results in an extra log, I wanted the option to customize that particular log without adding another one. Thanks a lot for the reply, we're gonna stick with this middlewares approach for now. Have a nice one!

PR #684, opened 2020-02-26
Use published module as HR time source

oliversalzburg:
This makes it a lot easier to generate tracing data that is compatible with moleculer. Other trace producers can just use performance-now and their desired tracing client to publish spans without even knowing about moleculer.

### DIFF
--- a/src/tracing/now.js
+++ b/src/tracing/now.js
@@ -13,4 +13,4 @@ function now() {
 const loadNs = now();
 const loadMs = Date.now();
 
-module.exports = () => loadMs + now() - loadNs;
+module.exports = () => loadMs + loadNs;


icebob:
Why do you think better to add a new dependency to the project instead of using the solution directly? The internal code is based on this lib. By the way, I struggle to minimalize the number of dependencies.

oliversalzburg:
@icebob Sorry about the hastly written PR. I was troubleshooting tracing issues and then got pulled to another project.

No matter what I tried, my spans I was submitting from a non-Moleculer context would never line up with the Moleculer spans and I couldn't figure it out. I assumed the time source in Moleculer was unreliable, even though it looked exactly like the code in performance-now. The inability to fully control this aspect of the tracing was a bit frustrating at the time.

However, I have since realized that the issue root cause for the issue is truely within the Moleculer source:
https://github.com/moleculerjs/moleculer/blob/master/src/tracing/now.js#L16

You see here, you are first adding `now()` to `loadMs` and then you subtract `loadNs`, which was just filled with `now()`. So the only thing added to the milliseconds is the tiny amount of microseconds between the two `now()` calls.

The "correct" implementation should be `module.exports = () => loadMs + loadNs;` and then the spans also line up across all my processes.

Funny enough, the change I proposed in this PR is worse, because it doesn't have the milliseconds part at all.

You're propably right for trying to keep dependencies to a minimum. If the Moleculer implementation could be adjusted to fix the above-mentioned issue, that'd be a lot better. I will probably update this PR with that change.

oliversalzburg:
Upon further review, I think I completely misread that. Things were going wrong for unrelated reasons. Sorry about that.

PR #697, opened 2020-03-17
Sanitize utils.safetyObject output from big objects

akazakou:
## :memo: Sanitize utils.safetyObject output from big objects

In some cases, we should pass into service `schemaMods` parameters a big amount of data. To example: some instances of [SharedArrayBuffer](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) class with data more than 100MB. In that case on service starting Moleculer.js tries to post that data into output/databus and it causes performance in `utils.safetyObject` to convert this amount of data into JSON.

### :dart: Relevant issues
- No related issues

### :gem: Type of change

### :scroll: Example code
```js
    const broker = new ServiceBroker({ logLevel: "debug" });
    broker.createService(Service, {
        name: 'maxmind',
        settings: {
            service: {
                database: new Buffer(Array.from({ length: 100 * 1024 * 1024 })),
            },
        },
    });
    broker.start();
``` 

Example of code to reproduce that issue provided in "Example code" section

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 # [Unreleased](https://github.com/moleculerjs/moleculer/compare/v0.14.3...v0.14.4)
 
 ## Changes
+- add `maxSafeObjectSize` to service broker options. Fixes [#697](https://github.com/moleculerjs/moleculer/pull/697)
 - add `stop` method to tracing exporters. Fixes [#689](https://github.com/moleculerjs/moleculer/issues/689)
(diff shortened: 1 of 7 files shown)


icebob:
On the other hand, the default value should be null & handle as infinite otherwise it causes breaking change.

akazakou:
@icebob Changes from ed821e7 commit
* Moved `maxSafeObjectSize` to service broker options. The default value for broker options now is `null` - checks will be disabled
* Updated unit tests to support new broker options in exists tests. 
* Added changes in CHANGELOG.md

PR #703, opened 2020-03-22
Adding MoleculerRepl type definition.

Wallacy:
Sorry about that, just forgot the repl return on type definition.

Now i think is complete.

Edit: For some reason Travis-CI did not install @type/vorpal. For this reason i just copy the type definition. Vorpal is not maintained anymore, and changes will be from MoleculerRepl not Vorpal. Also, Moleculer-repl uses moleculerjs/vorpal, not the original project.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,5 @@
 import { EventEmitter2 } from "eventemitter2";
+import Vorpal from "vorpal";
 
 declare namespace Moleculer {
 	/**
@@ -12,6 +13,7 @@ declare namespace Moleculer {
 	 *     type Promise<T> = Bluebird<T>;
 	 *   }
 	 */
+
 	type GenericObject = { [name: string]: any };
 
 	type LogLevels = "fatal" | "error" | "warn" | "info" | "debug" | "trace";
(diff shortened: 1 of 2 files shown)


icebob:
@shawnmcknight could you review it, please?

index.d.ts
```diff
@@ -1341,6 +1347,85 @@ declare namespace Moleculer {
 	const CIRCUIT_CLOSE: string;
 	const CIRCUIT_HALF_OPEN: string;
 	const CIRCUIT_OPEN: string;
+
+	class Vorpal {
```
shawnmcknight: @icebob Is it fair to suggest that moleculerjs/vorpal will now be maintained?  If so, it would probably be best to add these types there, cut a new release of that package, and then bump the package.json as part of this PR.  In that case, the new types would be available to both places.

icebob: I don't say it will be maintained because the source code is sophisticated, it was really tough to solve the last vulnerabilities issue, as well. I don't think we are able to add any other changes to this code. Of course, a type definition file can be easier, but I think Wallacy's definition is just a part of the full vorpal types, right?

Wallacy: I can put this on the moleculer/vorpal, also on molecular-repl, but i just put here because there's no previous type definition on those projects. This is a full vorpal definition by the way.

icebob: tomorrow I want to release a new moleculer version. @shawnmcknight if you said that the ts definition is correct, I will merge it & publish and in the future we will able to move it to the relevant repo, as well.

Wallacy: As a Typescript developer, this is a very straight forward definition anyway.

PR #715, opened 2020-04-09
Safely handle disconnected state for transporters

Nathan-Schwartz:
#### Issue
@icebob found that when connection is lost, some transporters queue messages in memory and send them when a connection is recovered. This is useful in some cases, but for Moleculer it can lead to queueing up many heartbeat packets and potentially requests.

#### Solution
If the connection is lost, BrokerDisconnectedError will be thrown for user-initiated packets when transporter disconnects (PING, REQ, EVENT). For internal packets (HEARTBEAT, INFO, etc) we will simply not send and no error will be thrown.

If the disconnection is long enough, heartbeats will be skipped and the remote nodes will be removed from the registry. After this happens, actions will start failing with ServiceNotAvailableError and RequestRejectedError like normally.

### Second commit
Address AMQP reconnection issue. See commit for more details.

### Third commit
Make sure `ServiceBroker.emit()` always returns a promise.

### :gem: Type of change

- [ ] Ran two nodes. One called actions on, emitted events to, and pinged the other node. Restarted message broker and confirmed that errors were thrown for the correct packets.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -61,6 +61,20 @@ class MoleculerRetryableError extends MoleculerError {
 }
 
 /**
+ * Moleculer Error class for Broker disconnections which is retryable.
+ *
+ * @class MoleculerServerError
+ * @extends {MoleculerRetryableError}
+ */
+class BrokerDisconnectedError extends MoleculerRetryableError {
+	constructor() {
+		super("The broker's transporter has disconnected. Please try again when a connection is reestablished.", 502, "BAD_GATEWAY");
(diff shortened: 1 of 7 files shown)


src/errors.js
```diff
@@ -60,6 +60,15 @@ class MoleculerRetryableError extends MoleculerError {
 	}
 }
 
+
+class BrokerDisconnectedError extends MoleculerRetryableError {
```
icebob, line 64: Please add tests for this new Error, as well into https://github.com/moleculerjs/moleculer/blob/master/test/unit/errors.spec.js

PR #718, opened 2020-04-16
Add getServiceList, missing in typescript definitions

HazemKhaled:
## :memo: Description

We are using this function in openapi.json auto-generate

### :gem: Type of change

### :scroll: Example code
```js
this.broker.registry.getServiceList({ withActions: true }).forEach(this.logger.info);
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1328,6 +1328,8 @@ declare namespace Moleculer {
 		services: any;
 		actions: any;
 		events: any;
+
+		getServiceList({withActions: bool}): ServiceSchema[]
 	}
 
 	class AsyncStorage {
--- a/index.d.ts
+++ b/index.d.ts
@@ -476,7 +476,7 @@ declare namespace Moleculer {
 		disconnected(): void;
 	}
 
-	class Context<P = unknown, M extends object = {}> {
+	class Context<P = unknown, M extends object = {[key: string]: any}> {
 		constructor(broker: ServiceBroker, endpoint: Endpoint);
 		id: string;
 		broker: ServiceBroker;


icebob:
Something is wrong with d.ts in Travis.

Wallacy:
I dont think this is the right signature:

```typescript
getServiceList({withActions: bool}): ServiceSchema[]
```

Per action-catalog:
https://github.com/moleculerjs/moleculer/blob/master/src/registry/action-catalog.js#L122

The options param has:
```typescript
{onlyLocal = false, onlyAvailable = false, skipInternal = false, withEndpoints = false}
```

So the best approach is:

```typescript
	interface ActionCatalogListOptions {
		onlyLocal:boolean = false;
		onlyAvailable:boolean = false;
		skipInternal:boolean = false;
		withEndpoints:boolean = false;
	}

	class ServiceRegistry {
		broker: ServiceBroker;
		metrics: MetricRegistry;
		logger: LoggerInstance;

		opts: BrokerRegistryOptions;

		StrategyFactory: BaseStrategy;

		nodes: any;
		services: any;
		actions: any;
		events: any;

		getServiceList(opts?: ActionCatalogListOptions): ServiceSchema[]
	}

```

This also will solve the Travis-ci problem;

icebob:
@Wallacy the `ActionCatalogListOptions ` correct, but I don't like to define more detail of the `ServiceRegistry` because it's an internal class, and the methods are changing sometimes.

Wallacy:
So, in this case better just keep only:

```javascript
{onlyLocal = false, onlyAvailable = false, skipInternal = false, withEndpoints = false}
```
From: https://github.com/moleculerjs/moleculer/blob/master/src/registry/action-catalog.js#L122

Wallacy:
@icebob This was done on #735 too btw;

icebob:
Great, thanks!

Issue #261, opened 2018-05-07
WebSockets Transport Layer


Wallacy:
Im working on a WebSockets Transport Layer (just started) for Moleculer, and i'm open this issue to discuss.

I don't know if anyone has this need to, but in may case, i need to push a real-time data to the end user.

The idea is not make a complete pub/sub, wrapper like socket.io, but a complement to API Gateway (moleculer-web) at some point.

First i'm planing to make a regular WebSockets transport for the server side, after that work go to the client side, maybe implementing a lite version of the moleculer broker to be used from the browser.

In the future im planning adding a WebRTC/SCTP transport too. (just a quick [reference ](http://blog.brkho.com/2017/03/15/dive-into-client-server-web-games-webrtc/)for why)

Wallacy:
Nice, i missed this project. So, i'm not alone xD

BTW:
Will be nicer to call moleculer-web as moleculer-gateway, and then make the "rest" handler moleculer-rest, and plug into gateway?
In this way we can easily insert a molecular-ws, molecular-soap or other client side communication mechanism.

kevinsegal:
I've found that moleculer runs very well with Primus (with uws under the hood); and I've been working on a cluster-compatible version based on Socketcluster which also has been reliable so far.  In terms of using this as a transporter internally, would it really lead to a noticeable difference in speed for the end-user, vs NATS or the other transporters currently available ?

Wallacy:
I did not finish yet. I’m also working on service discovery for the websocket, but my early tests is very close to the TCP implementation, so will be fast than NATS.

For now, im only trying to create a broken that at some point will be able to communicate directly to the browser without a 3rd party project like Primus.

After that a will put on the table some things that can make a difference for some people. (Service Discovery is already one)

icebob:
Any updates about it? Are you working on it yet?

Wallacy:
Hello,

I did a simple version, but i got a lot of work last 3 weeks and did not have a change to finish. Sorry about the delay, i plan to rollout next week. Sorry about that.

Wallacy:
Hello,

Probably my week is measured on mars or jupiter calendar! I will try again this week! ;)

icebob:
Any news? :)

Wallacy:
Sorry, i need more time. Too busy! But you can close the issue if want. When i finish i will make the P.R in any case.

;)

Wallacy:
Finally i get the time to do this!

[...]

PR #731, opened 2020-05-17
New Discoverer modules (Local, Redis, etcd3)

icebob:
## New Discoverer module
The Discoverer is a new built-in module in Moleculer framework. It's responsible for that all Moleculer nodes can discover each other and check them with heartbeats. In previous versions, it was an integrated module inside `ServiceRegistry` & `Transit` modules. In this version, the discovery logic has been extracted to a separated built-in module. It means you can replace it to other built-in implementations or a custom one. The current discovery & heartbeat logic is moved to the `Local` Discoverer. 

Nevertheless, this version also contains other discoverers, like Redis & etcd3 discoverers. Both of them require an external server to make them work. 
One of the main advantages of the external discoverers, the node discovery & heartbeat packets don't overload the communication on the transporter. The transporter transfers only the request, response, event packets.
By the way, the external discoverers have some disadvantages, as well. These discoverers can detect lazier the new and disconnected nodes because they scan the heartbeat keys periodically on the remote Redis/etcd3 server. The period of checks depends on the `heartbeatInterval` broker option.

>The Local Discoverer (which is the default) works like the discovery of previous versions, so if you want to keep the original logic, you'll have to do nothing.

>Please note the TCP transporter uses Gossip protocol & UDP packets for discovery & heartbeats, it means it can work only with Local Discoverer.

### Local Discoverer
It's the default Discoverer, it uses the transporter module to discover all other moleculer nodes. It's quick and fast but if you have too many nodes (>100), the nodes can generate a lot of heartbeat packets which can reduce the performance of request/response packets.

**Example**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: "Local"
    }    
}
```

**Example with options**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: {
            type: "Local",
            options: {
                // Send heartbeat in every 10 seconds
                heartbeatInterval: 10,

                // Heartbeat timeout in seconds
                heartbeatTimeout: 30,

                // Disable heartbeat checking & sending, if true
                disableHeartbeatChecks: false,

                // Disable removing offline nodes from registry, if true
                disableOfflineNodeRemoving: false,

                // Remove offline nodes after 10 minutes
                cleanOfflineNodesTimeout: 10 * 60
            }
        }
    }    
}
```

This Discoverer uses a [Redis server](https://redis.io/) to discover Moleculer nodes. The heartbeat & discovery packets are stored in Redis keys. Thanks to Redis key expiration, if a node crashed or disconnected unexpectedly, the Redis will remove its heartbeat keys from the server and other nodes can detect it.

**Example to connect local Redis server**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: "Redis"
    }    
}
```

**Example to connect remote Redis server**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: "redis://redis-server:6379"
    }    
}
```

**Example with options**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: {
            type: "Redis",
            options: {
                redis: {
                    // Redis connection options.
                    // More info: https://github.com/luin/ioredis#connect-to-redis
                    port: 6379,
                    host: "redis-server",
                    password: "123456",
                    db: 3
                }

                // Serializer
                serializer: "JSON",

                // Full heartbeat checks. It generates more network traffic
                // 10 means every 10 cycle.
                fullCheck: 10,

                // Key scanning size
                scanLength: 100,

                // Monitoring Redis commands
                monitor: true,
                
                // --- COMMON DISCOVERER OPTIONS ---

                // Send heartbeat in every 10 seconds
                heartbeatInterval: 10,

                // Heartbeat timeout in seconds
                heartbeatTimeout: 30,

                // Disable heartbeat checking & sending, if true
                disableHeartbeatChecks: false,

                // Disable removing offline nodes from registry, if true
                disableOfflineNodeRemoving: false,

                // Remove offline nodes after 10 minutes
                cleanOfflineNodesTimeout: 10 * 60
            }
        }
    }    
}
```

>To be able to use this Discoverer, install the `ioredis` module with the npm install ioredis --save command.

>Tip: To further network traffic reduction, you can use MsgPack/Notepack serializers instead of JSON.

This Discoverer uses an [etcd3 server](https://etcd.io/) to discover Moleculer nodes. The heartbeat & discovery packets are stored in the server. Thanks to etcd3 [lease](https://etcd.io/docs/v3.4.0/learning/api/#lease-api) solution, if a node crashed or disconnected unexpectedly, the etcd3 will remove its heartbeat keys from the server and other nodes can detect it.

**Example to connect local etcd3 server**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: "Etcd3"
    }    
}
```

**Example to connect remote etcd3 server**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: "etcd3://etcd-server:2379"
    }    
}
```

**Example with options**
```js
// moleculer.config.js
module.exports = {
    registry: {
        discoverer: {
            type: "Etcd3",
            options: {
                etcd: {
                    // etcd3 connection options.
                    // More info: https://mixer.github.io/etcd3/interfaces/options_.ioptions.html
                    hosts: "etcd-server:2379",
                    auth: "12345678"
                }

                // Serializer
                serializer: "JSON",

                // Full heartbeat checks. It generates more network traffic
                // 10 means every 10 cycle.
                fullCheck: 10,
                
                // --- COMMON DISCOVERER OPTIONS ---

                // Send heartbeat in every 10 seconds
                heartbeatInterval: 10,

                // Heartbeat timeout in seconds
                heartbeatTimeout: 30,

                // Disable heartbeat checking & sending, if true
                disableHeartbeatChecks: false,

                // Disable removing offline nodes from registry, if true
                disableOfflineNodeRemoving: false,

                // Remove offline nodes after 10 minutes
                cleanOfflineNodesTimeout: 10 * 60
            }
        }
    }    
}
```

>To be able to use this Discoverer, install the `etcd3` module with the npm install etcd3--save command.

>Tip: To further network traffic reduction, you can use MsgPack/Notepack serializers instead of JSON.

### Custom Discoverer
You can create your custom Discoverer. We recommend to copy the source of Redis Discoverer and implement the necessary methods.


Wallacy:
Wow, thats Nice!

I playing on something similar to this myself; I was planning to add two new transporters (_Pure UDP_ and [WebSockets](https://github.com/moleculerjs/moleculer/issues/261) that I'm in debt ); And i was looking a way to reuse the Gossip/Discovery from TCP;

This PR looks great! if you want i can push a upgraded `index.d.ts` tomorrow.

icebob:
@Wallacy updating index.d.ts would be great! I totally forgot about it. Thanks in advance. I merge it, so please send it with a new PR to the master.
Thanks in advance.

_By the way, I plan to make a HTTP transporter because it can work with external discoverers._

wuyuedefeng:
Why not support consul?

ramyak-mehra:
Hey I was planning to use the reds discoverer on prod, but saw a warning that its still experimental. Are there any known bugs on using this? Is there a list or roadmap regarding it.  Maybe I could also pick those up.

icebob:
We don't know of any bugs. But there is a limitation in the current Discoverer implementation. In the case of many Moleculer nodes (~500), the startup time of nodes will be high (it parses and processes all downloaded nodes' info).

So I plan to rewrite and optimize the logic of discoverers in later versions.

PR #735, opened 2020-05-19
Updating index.d.ts

Wallacy:
Adding BrokerRegistryOptions > `discoverer` > `RegistryDiscovererOptions`

I took the liberty of adding the correct definition of  #718

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -705,6 +705,12 @@ declare namespace Moleculer {
 		strategy?: Function | string;
 		strategyOptions?: GenericObject;
 		preferLocal?: boolean;
+		discoverer?: RegistryDiscovererOptions | string;
+	}
+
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -705,6 +705,12 @@ declare namespace Moleculer {
 		strategy?: Function | string;
 		strategyOptions?: GenericObject;
 		preferLocal?: boolean;
+		discoverer?: RegistryDiscovererOptions | string;
```
icebob, line 708: It can be an instance of `BaseDiscoverer`, as well

Wallacy: Yes you are right, i will make the changes. 
I tried to keep it very simple, but on second thought it is better to make it more complete as in 'Strategies';

Wallacy:
Okay; I update the definition;

Issue #741, opened 2020-05-22
[lodash] Remove usage of _.isPlainObject


abdavid:
See #433 for context.

**Describe the solution you'd like**
Remove usage of `_.isPlainObject` throughout Moleculer. 

**Describe alternatives you've considered**
Should be replaceable by `Object.prototype.toString.call(o) ==="[object Object]"`

**Additional context**
https://github.com/moleculerjs/moleculer/search?q=_.isPlainObject&unscoped_q=_.isPlainObject

Wallacy:
```javascript
function isPlainObject(o) {
    return Object.prototype.toString.call(o) ==="[object Object]"
}

function Foo() {
   this.a = 1
 }

$ isPlainObject(new Foo)
true
```

If the ideia is match _.isPlainObject

```javascript
function isPlainObject(obj) {
  return obj!=null ? Object.getPrototypeOf(obj)===Object.prototype || Object.getPrototypeOf(obj)===null : false
}
```
```javascript
$ isPlainObject(null)
false
$ isPlainObject(new Foo)
false
$ isPlainObject([1, 2, 3])
false
$ isPlainObject({ 'x': 0, 'y': 0 })
true
$ isPlainObject(Object.create(null))
true
```

abdavid:
Nice! Maybe you would want to create a PR for that? 👍

Wallacy:
Okay, i can do that.

Issue #740, opened 2020-05-22
[lodash] Remove usage of _.isString


abdavid:
See #433 for context.

**Describe the solution you'd like**
Remove usage of `_.isString` throughout Moleculer. 

**Describe alternatives you've considered**
Should be replaceable with a `Object.prototype.toString.call(str) === "[object String]"`

**Additional context**
https://github.com/moleculerjs/moleculer/search?q=_.isString&unscoped_q=_.isString

Wallacy:
Why not a simple `typeof a_string === 'string'` ?

abdavid:
That will suffice also, I was merely contemplating a reusable objectType detetion pattern for all the other `isX` utilities, but since we didnt go that way a regular `typeof x === "string"` will certainly do the trick.

PR #760, opened 2020-05-25
Fix for _.isString and _.isPlainObject

Wallacy:
This fixes: #740, and fix: #741

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -73,11 +73,11 @@ const utils = {
 	},
 
 	isString(s) {
-		return typeof s === "string";
+		return typeof s === "string" || s instanceof String;
 	},
 
 	isObject(o) {
-		return o !== null && typeof o === "object";
+		return o !== null && typeof o === "object" && !(o instanceof String);
 	},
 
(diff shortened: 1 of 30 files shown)


src/utils.js
```diff
@@ -73,10 +72,18 @@ const utils = {
...
+		return typeof s === "string";
+	},
+
 	isObject(o) {
 		return o !== null && typeof o === "object";
 	},
 
+	isPlainObject(o) {
+		return o !=null ? Object.getPrototypeOf(o) === Object.prototype || Object.getPrototypeOf(o) === null : false;
+	},
```
icebob, line 85: Please add unit tests for them.

Wallacy: Ops... I forget that.

Wallacy: Done... 

Doing the tests a observed the isObject test was incomplete, then i added more tests. Doing that i observed that object-wrapped strings was not handled and i added a fix to that;

icebob: Great, thanks!

PR #761, opened 2020-05-26
configurable interval in waitForServices function

ewen-marechal:
First, many thanks to @icebob and the community for this amazing framework. This is my first contribution to an open-source project, so I hope that I've respected the workflow.

## :memo: Description
In noticed that moleculer could take several seconds(1) to startup due to the dependency interval that is hard-coded to 1000 ms. Using the following ServiceFactory, I managed to reduce the startup time a lot when dependencies are involved.
```js
class FastService extends Service {
  constructor(broker, schema) {
    super(broker, schema)
  }
  waitForServices(serviceNames, timeout, interval) {
    return this.broker.waitForServices(serviceNames, timeout, this.settings.$dependencyInterval || 1000, this.logger)
  }
}
```
This pull request slightly changes the code of waitForServices for both service and service-broker, in order to allow the configuration of the interval parameter.

(1) `~1000 * n ms`, where n is the depth of the dependency chain

### :gem: Type of change

I add a test in service.spec "should call waitForServices if dependencies are defined & $dependencyInterval". It tests that adding $dependencyInterval in service.settings calls waitForServices with correct arguments

### DIFF
--- a/src/service-broker.js
+++ b/src/service-broker.js
@@ -104,6 +104,7 @@ const defaultOptions = {
 
 	internalServices: true,
 	internalMiddlewares: true,
+	dependencyInterval: 1000,
 
 	hotReload: false,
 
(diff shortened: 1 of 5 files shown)


ewen-marechal:
I added 2 tests in service-broker.spec.js
- one is for testing that the `dependencyInterval` option is used properly
- one is for testing that a shorter `dependencyInteval` allows to avoid a timeout in a dependency chain.

Service.waitForServices will call Broker.waitForServices with `this.broker.options.dependencyInterval` when `this.settings.$dependencyInterval` is falsy (rather than 0). I changed tests in service.spec.js accordingly

PR #763, opened 2020-05-27
async-storage unit test

IstenesImre:
## :memo: Description

Small unit test, if you find it good, I would write even more.


icebob:
Thank you! More tests are always good ;)

PR #775, opened 2020-06-19
replace slave with replica

stephanos:
## :memo: Description

Removes antiquated word "slave" and replaces it with "replica".

Doesn't touch the word "master" since it might break things (dependency `cluster` defines `isMaster`).

### :gem: Type of change

Since I've only renamed things, I haven't added any additional tests.

- [ ] I have added tests that prove my fix is effective or that my feature works **[N/A]**
- [x] New and existing unit tests pass locally with my changes
- [ ] I have commented my code, particularly in hard-to-understand areas **[N/A]**

### DIFF
--- a/src/strategies/latency.js
+++ b/src/strategies/latency.js
@@ -15,7 +15,7 @@ const BaseStrategy = require("./base");
  *
  * Since Strategy can be instantiated multiple times, therefore,
  * we need to have a "master" instance to send ping, and each
- * individual "slave" instance will update their list dynamically
+ * individual "replica" instance will update their list dynamically
  *
  * These options can be configured in broker registry options:
  *
@@ -76,10 +76,10 @@ class LatencyStrategy extends BaseStrategy {
 			this.broker.localBus.on("$broker.stopped", () => this.brokerStopped = true);
 		} else {
 			// remove node if we are told by master
-			this.broker.localBus.on("$node.latencySlave.removeHost", this.removeHostLatency.bind(this));
+			this.broker.localBus.on("$node.latencyReplica.removeHost", this.removeHostLatency.bind(this));
 		}
 
-		this.broker.localBus.on("$node.latencySlave", this.updateLatency.bind(this));
+		this.broker.localBus.on("$node.latencyReplica", this.updateLatency.bind(this));
 	}
 
 	// Master
@@ -129,7 +129,7 @@ class LatencyStrategy extends BaseStrategy {
 
 		const avgLatency = info.historicLatency.reduce((sum, latency) => sum + latency, 0) / info.historicLatency.length;
 
-		this.broker.localBus.emit("$node.latencySlave", {
+		this.broker.localBus.emit("$node.latencyReplica", {
 			hostname: node.hostname,
 			avgLatency: avgLatency
 		});
@@ -173,17 +173,17 @@ class LatencyStrategy extends BaseStrategy {
 
 		if (info.nodeList.length == 0) {
 			// only remove the host if the last node disconnected
-			this.broker.localBus.emit("$node.latencySlave.removeHost", node.hostname);
+			this.broker.localBus.emit("$node.latencyReplica.removeHost", node.hostname);
 			this.hostMap.delete(node.hostname);
 		}
 	}
 
-	// Master + Slave
+	// Master + Replica
 	updateLatency(payload) {
 		this.hostAvgLatency.set(payload.hostname, payload.avgLatency);
 	}
 
-	// Slave
+	// Replica
 	removeHostLatency(hostname) {
 		this.hostAvgLatency.delete(hostname);
 	}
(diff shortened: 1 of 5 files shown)


src/strategies/latency.js
```diff
@@ -173,17 +173,17 @@ class LatencyStrategy extends BaseStrategy {
 
 		if (info.nodeList.length == 0) {
 			// only remove the host if the last node disconnected
-			this.broker.localBus.emit("$node.latencySlave.removeHost", node.hostname);
+			this.broker.localBus.emit("$node.latencyReplica.removeHost", node.hostname);
```
stephanos, line 176: from what I could tell this is an internal identifier, not part of the public API. but I might be wrong?

src/strategies/latency.js
```diff
@@ -173,17 +173,17 @@ class LatencyStrategy extends BaseStrategy {
...
 			this.hostMap.delete(node.hostname);
 		}
 	}
 
-	// Master + Slave
+	// Master + Replica
```
d0whc3r, line 181: if you change "Slave" for "Replica" you could also change "Master"

d0whc3r: https://www.vice.com/en_us/article/8x7akv/masterslave-terminology-was-removed-from-python-programming-language

intech:
@stephanos dear Stephan,
I don’t know how good a engineer you are, but your PR will break many dependents.

> A right is not what someone gives you; it's what no one can take from you.
Ramsey Clark.

You should not take away our freedom of engineering under the guise of upholding the rights of another social group.
This change will not help change historical events. We must draw conclusions from mistakes and never repeat them again.

stephanos:
> I don’t know how good a(n) engineer you are

How kind of you.

> but your PR will break many dependents

Well, I couldn't tell from the code or the docs whether it was a breaking change. As far as I could tell, it's not part of the publicly documented API. Hence my question. Maybe the docs need improvement.

> You should not take away our freedom of engineering under the guise of upholding the rights of another social group.

I'm not "taking away your freedom" since you are in charge of the library. Just say no. I'm raising awareness that this is an outdated word. The choice is still yours. The word isn't _outlawed_, but if there's a replacement, why not use that, all things being equal?

> This change will not help change historical events.

Obviously, only time traveling would.

> We must draw conclusions from mistakes and never repeat them again

:-? What's _your_ conclusion?

Nathan-Schwartz:
Hey @stephanos, thanks for investing time into the project!

I understand if you want to close this PR, but just so you know @intech doesn't officially speak on behalf of Moleculer. (I don't either). In my opinion, this decision is really up to @icebob.

Obviously this is a divisive topic, but I think insinuating that you are a bad engineer was unnecessary and and impolite. I'm sorry this was your first experience contributing to Moleculer.

Breaking changes happen from time-to-time, and at this stage of the project, they happen frequently enough that we ship them in minor version changes. The fact that this is a breaking change doesn't, in my opinion, warrant automatically dismissing this effort.

Re: whether or not this change should be made, I'll remain neutral for now. I see some valid points on both sides.

intech:
Oh, I didn’t think about offending @stephanos somehow. Please excuse me if it was written rudely.
I wanted to say that Stefan did this PR without understanding how the product works.

@Nathan-Schwartz you are right, I am not the face of moleculer, and it is correct that no one is him. This is a community and open source project for everyone by MIT license. I don’t think it’s right to talk about it as someone’s property.
If this project belongs to a particular person, then this should be written in a license so that people and companies that use this code understand that their risks depend on the decision of one person, not the community.
During the existence of the project, I was convinced that this is not so and I believe that it will never be so.

P.S.: The example with 'ruskie' / 'moskal' will be interesting only to people who do not live in Russia. In Russia there are more important problems and no one will pay attention to this.
I concluded that we should be above all this. Social groups have always been, are and will be even greater; organizing a struggle for the sake of struggle is not beneficial for anyone.
I am for peace and the exchange of experience between all people on the planet without separation on any basis.

Peace!

Nathan-Schwartz:
Thanks for clarifying @intech 🙂
I understand. I didn't meant to bring up governance or ownership, my goal was to let @stephanos know he didn't need to close the PR and we can keep discussing it. (Sorry for leading us on a tangent)

I have two questions for general discussion:
1. If we decide to change terminology here, what is the best replacement? 
a) primary/replica
b) leader/follower
c) something else

2. This PR affects the name of one local event for one non-default LB strategy. The event is primarily for internal use and it isn't documented on the doc site. Is my understanding correct?

src/strategies/latency.js
Nathan-Schwartz: I agree (If it isn't a much bigger breaking change)

src/strategies/latency.js
Nathan-Schwartz: I believe it's meant for internal use, but devs can subscribe to it if they want to.

PR #778, opened 2020-06-24
Constrain service settings in type definitions

shawnmcknight:
## :memo: Description

Currently a TS consumer is allowed to provide a generic to describe the `ServiceSettingSchema` for both the `ServiceSchema` and the `Service`.  This will result in a strongly typed `settings` property object which then can guarantee type safety on `this.settings` in the service rather than a free-form `any` result being returned from all properties.

However, the generic is not constrained and thus use of this can cause the loss of the built-in `$` prefixed values that are provided by default with moleculer.  In practicality, any override of the generic needs to extend an interface which includes those built-in types.

This PR splits those built-in properties into a `ServiceSettingBaseSchema`.  The existing `ServiceSettingSchema` then extends from that base but adds an index signature to allow an object with any type for its properties (that is, no type safety, but flexible).  The generic still defaults from this flexible `ServiceSettingSchema` so consumers who were not providing an override for the generic will see no change.  However, a constraint was added to the generic that it must extend `ServiceSettingBaseSchema`.  This means that any generic provided must extend that interface.  This will guarantee that any provider of an override to the generic will have access to the built-in properties as well.

This is a minor breaking change to any consumers currently providing an override for this generic which does not extend `ServiceSettingSchema` or provide types for the built-in properties.  The fix is merely to add `extends ServiceSettingBaseSchema` to those existing interfaces.

### :gem: Type of change

### :scroll: Example code
```ts
import { ServiceSettingBaseSchema } from 'moleculer';

interface FooServiceSettingSchema extends ServiceSettingBaseSchema {
  foo: number;
  bar: string;
}

class FooService extends Service<FooServiceSettingSchema > {
  public constructor(broker: ServiceBroker) {
    super(broker);

    this.parseServiceSchema({
      settings: {
        foo: 1337,
        bar: 'bar',
      },
      ...
    });
  }
...
}
``` 

Ran in local environment with services that both defined overrides for the generic and those that did not.  All types worked properly once the interface was extended.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -543,12 +543,15 @@ declare namespace Moleculer {
 		static create(broker: ServiceBroker): Context;
 	}
 
-	interface ServiceSettingSchema {
+	interface ServiceSettingBaseSchema {
 		$noVersionPrefix?: boolean;
 		$noServiceNamePrefix?: boolean;
 		$dependencyTimeout?: number;
 		$shutdownTimeout?: number;
 		$secureSettings?: Array<string>;
+	}
+
+	interface ServiceSettingSchema extends ServiceSettingBaseSchema {
 		[name: string]: any;
 	}
 
@@ -614,7 +617,7 @@ declare namespace Moleculer {
 		version?: string | number;
 	}
 
-	interface ServiceSchema<S = ServiceSettingSchema> {
+	interface ServiceSchema<S extends ServiceSettingBaseSchema = ServiceSettingSchema> {
 		name: string;
 		version?: string | number;
 		settings?: S;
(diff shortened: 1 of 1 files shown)


icebob:
The question is how we handle the TS-only breaking changes. If it's same as API breaking changes for JS users, then we can release it only in the next 0.15 version.

shawnmcknight:
> The question is how we handle the TS-only breaking changes. If it's same as API breaking changes for JS users, then we can release it only in the next 0.15 version.

It's a no-impact change for JS as this is type only.  It's always challenging to figure out how best to approach this type of thing.  The existing type is technically wrong because the `settings` object would not match the provided generic unless it extends the built-in.

We could split the difference here.  Introduce the new base type, but not bring in the constraint.  Existing consumers would see no impact at all, although their settings object would continue to be incorrect.  We could then introduce the constraint in 0.15.

With that said, I run into TS definition issues _all the time_ when updating package versions even when its not flagged as a semver breaking change.  That might not be the best answer, but its hard to semver both JS and TS simultaneously.  I'm okay with playing this one cautious and following up at a later date though.

PR #786, opened 2020-07-17
Remove Zipkin v1 annotations

kthompson23:
#385  :memo: Description
This PR updates the Zipkin Exporter payload to conform to the V2 API. In the V2 API, annotations should be used for interesting events like errors or to indicate latency. The short codes like 'SS' and 'SR' are no longer used to infer the kind of span. Instead, the `kind` field should be used.

Removing these annotations resolves an issue we are experiencing when using the Honeycomb.io exporter in the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib) where these annotations are interpreted as "span events". A span event is a structured log without a duration that is associated with an actual span. So for things like errors that are also put into annotations, having the span event within the actual span is meaningful and interesting. Whereas the 'SS' and 'SR' annotations duplicate the start span timestamp and duration and add a lot of noise to our traces.

Please see the [annotation](https://zipkin.io/zipkin-api/#) section:
>Associates an event that explains latency with a timestamp.
Unlike log statements, annotations are often codes. Ex. “ws” for WireSend

>Zipkin v1 core annotations such as “cs” and “sr” have been replaced with
Span.Kind, which interprets timestamp and duration.

I also updated the span kind to be SERVER since I believe this is more accurate than CLIENT.

Please see the [span](https://zipkin.io/zipkin-api/#) section
> SERVER
    timestamp is the moment a client request was received. (in v1 “sr”)
    duration is the delay until a response was sent or an error. (in v1 “ss”-“sr”)

### :gem: Type of change

- [x] Bug fix. It's possible users of the Zipkin exporter were using these annotations but this data was already available as the span start time and duration.

### :scroll: Example code
```js

``` 

I ran our service stack using the Zipkin exporter and confirmed that the spans were emitted correctly.

### DIFF
--- a/src/tracing/exporters/zipkin.js
+++ b/src/tracing/exporters/zipkin.js
@@ -145,7 +145,7 @@ class ZipkinTraceExporter extends BaseTraceExporter {
 		const serviceName = span.service ? span.service.fullName : null;
 		const payload = {
 			name: span.name,
-			kind: "CONSUMER",
+			kind: "SERVER",
 
 			// Trace & span IDs
 			traceId: this.convertID(span.traceID),
@@ -155,10 +155,7 @@ class ZipkinTraceExporter extends BaseTraceExporter {
 			localEndpoint: { serviceName },
 			remoteEndpoint: { serviceName },
 
-			annotations: [
-				{ timestamp: this.convertTime(span.startTime), value: "sr" },
-				{ timestamp: this.convertTime(span.finishTime), value: "ss" },
-			],
+			annotations: [],
 
 			timestamp: this.convertTime(span.startTime),
 			duration: this.convertTime(span.duration),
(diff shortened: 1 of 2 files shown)


icebob:
Can it cause breaking change for Zipkin v1 users?

kthompson23:
This isn't a breaking change because Moleculer was always emitting Zipkin traces in the v2 JSON format. The v1 and v2 formats are incompatible so any users of the existing Moleculer Zipkin exporter were sending them to a v2 endpoint.

I think the change I made to the `kind` property from `CONSUMER` to `SERVER` is a bug fix. It is possible to query on annotations but I can't imagine anyone was querying on the "sr" and "ss" annotations because that information is easier to access via the span timestamp and duration.

Most v2 api consumers are probably ignoring these fields. For example, the NewRelic Zipkin consumer does not support annotations. Neither did the Honeycomb consumer we were using. The new OpenTelemetry collector that I am trying to switch to does support annotations but expects them to be things like errors. It does not expect the "sr", "ss", "cs", etc. codes to be there.

icebob:
Thanks the clarification & PR!

PR #789, opened 2020-07-21
Add optional global action and event tracing tags to apply for all actions and events

### LINKED ISSUE #784: Add ability to add global action tracing tags
I am using the Zipkin exporter to implement tracing. I would like to add a few fields from context.meta on all spans. Currently, as far as I can tell, the only way to add fields from context.meta is to add tracing options in each action.

We have a lot of services and actions so this would be time-consuming, repetitive, and possible forgotten in future actions that we create.

Would you accept a PR to add an optional option called `tags` to the Tracer options? This would apply to all actions and be override-able at the action level. 

Tracer options:
```js
tracing: {
	enabled: true,
	exporter: {
		type: 'Zipkin',
	},
	tags: {
		meta: ['app.id', 'user.email', 'workspace.name'],
	},
},
```

In the tracing middleware we'd do a merge like:
```js
// override global action tags with action tags
const mergedTags = { ...tracer.tags, ...opts.tags }
if (_.isFunction(mergedTags)) {
	const res = mergedTags.call(ctx.service, ctx);
	if (res)
		Object.assign(tags, res);

} else if (_.isPlainObject(mergedTags)) {
	if (mergedTags.params === true)
		tags.params = ctx.params != null ? Object.assign({}, ctx.params) : ctx.params;
	else if (Array.isArray(mergedTags.params))
		tags.params = _.pick(ctx.params, mergedTags.params);

	if (mergedTags.meta === true)
		tags.meta = ctx.meta != null ? Object.assign({}, ctx.meta) : ctx.meta;
	else if (Array.isArray(mergedTags.meta))
		tags.meta = _.pick(ctx.meta, mergedTags.meta);
```

kthompson23:
# :memo: Description

This PR adds a new tracing option called `tags`. Here the user may specify action and/or event tags that will apply to all actions and events. The options are identical to the local action and event tags meaning that the user can either blanket turn on/off params, meta, and response (action only). They can specify a custom function that accepts `ctx` and an optional `response` (action only), or an array of fields that will be used to pick properties off of the meta, params, or response (action only).

The following hierarchy applies.

1. Default behavior - all params are applied 
2. Global action/event tags or functions are applied
3. Local action/event tags or functions are applied

This means that the default behavior can be overridden globally and locally. Global tags can be overridden locally.

The actual algorithm is a shallow merge unless custom functions are specified. A shallow merge means that the user can specify meta, params, and response tags in both local actions/events and globally. The local tags will take precedence but if a user specified a global meta tag and only a local params tag they will both be applied. If both local meta tags and global meta tags are supplied then the local meta tags will be used.

Finally, the Typescript types were updated.

### :dart: Relevant issues
 Closes #784

### :gem: Type of change

### :scroll: Example code
```js
const tracingOptions = {
    enabled: true,
    events: true,
    exporter: 'Zipkin',
    tags: {
        action: {
            meta: ['app.id', 'user.email', 'workspace.name'],
            params: false,  // overrides default behavior of all params being adding as tags
            response: true,
        },
        event: (ctx) {
           return {
               caller: ctx.caller,
          }
       }
    }
},
``` 

These changes were run against our service stack and the zipkin exporter. I tried a combination of global and local tags and verified that they were being applied correctly.

I noticed that not all tests were running when running `npm run test:unit`. I updated the `unit` and `int` scripts to recurse down their respective directories.


kthompson23:
Thank you.

I will do a follow-up PR to the docs site.

intech:
@kthompson23 thanks for you PR, it was very useful for me.

icebob:
Hi @kthompson23 could you check this issue that what is the good behaviour?
https://github.com/moleculerjs/moleculer-web/issues/210#issuecomment-704819655

Issue #797, opened 2020-08-17
Trace spans should include the requestID as a tag to improve searchability


kthompson23:
In the Datadog, Zipkin, New Relic, and Jaeger exporters, the requestID is used as the basis for the traceId. However, in all implementations the requestID is run through a convert function to strip dashes and truncate it to 16 characters. In the Jaeger exporter it is further modified by converting to hex. This makes it very difficult to search for traces using the requestID because you must know how the exporter you are using modified the requestID to generate the traceId.

Moleculer should include the requestID as a tag in all spans so users can search for spans and traces using the requestID. In our environment we include the requestID in all logs and error messages. This will allow us to easily match up logs and errors with traces and spans.

icebob:
Good idea. Could you create a PR? I think we should add it into these parts:
https://github.com/moleculerjs/moleculer/blob/master/src/middlewares/tracing.js#L28
https://github.com/moleculerjs/moleculer/blob/master/src/middlewares/tracing.js#L146

kthompson23:
Sure, happy to do a PR.

Any issues with adding it in the `Context.startSpan` method? That way user defined tracing spans created using ctx would also get the requestID.

Something like:

```js
/**
 * Start a new child tracing span.
 *
 * @param {String} name
 * @param {Object?} opts
 * @returns {Span}
 * @memberof Context
 */
startSpan(name, opts = {}) {
	if (!opts.tags) {
		opts.tags = { requestID: this.requestID };
	}
	else if (opts.tags.requestID == null) {
		opts.tags.requestID = this.requestID;
	}

	let span;
	if (this.span) {
		span = this.span.startSpan(name, Object.assign({ ctx: this }, opts));
	} else {
		span = this.broker.tracer.startSpan(name, Object.assign({ ctx: this }, opts));
	}

	this._spanStack.push(span);
	this.span = span;

	return span;
}
```

icebob:
Awesome!

> Any issues with adding it in the Context.startSpan method? That way user defined tracing spans created using ctx would also get the requestID.

Ok, let's try it.

ehaselwanter:
This might help with what I want to achieve. I have traefik in front of moleculer. I can see traces in jaeger entering traefik, I can trace them **in** moleculer, but I just don't get it right to follow the request from the web down to one of the services ..

<img width="370" alt="image" src="https://user-images.githubusercontent.com/18671/94897129-505c6900-048f-11eb-8d59-3d96d8a3f40d.png">

[...]

PR #802, opened 2020-08-27
Add requestID tag to all action and event spans

kthompson23:
# :memo: Description

This PR updates all action and event spans to include the requestID. Traces will now be queryable by requestID. This is useful to us because we always return the requestID in the http header and log it with errors.

### :dart: Relevant issues
#797 

### :gem: Type of change

### :scroll: Example code
```js

``` 

I confirmed that spans created by the action middleware automatically included the requestID as a trace tag.


kthompson23:
If anyone feels strongly about not adding the requestID by default to all spans created using context I'll update the PR to only add it from the action middleware like you originally suggested. I think having the requestID present will be useful more often than not though.

kthompson23:
I reverted this and went with your original suggestion to add it to action and event tags in the middleware. I was thinking more about our custom spans and I realized we'd never query for a requestID **and** whatever tags we have in our custom spans. If users of custom spans want to add the requestID it is trivial to do so.

test/unit/middlewares/tracing.spec.js
```diff
@@ -1078,9 +1080,10 @@ describe("Test TracingMiddleware localAction", () => {
 							retries: 3,
 							timeout: 5
 						},
+						requestID: "request-id",
 						remoteCall: false
 					},
-					traceID: "tracer-trace-id",
```
icebob, line 1083: Why is it also changed?

kthompson23: When I originally did the global action/event tags I didn't realize that you were setting the requestID to null to test if `getCurrentTraceID` was called. Since the the tests for the global action/event tags have nothing to do with this behavior I thought this was a good opportunity to clean up these tests to reduce the number of assertions in them.

icebob (approved):
Thanks, good job!

PR #816, opened 2020-09-27
Fix: Add "locals" member for Context class in types definitions.

wiva-koikoi:
## :memo: Description

Add "locals" member for Context class in types definitions as it present in the context but not in its type definition.

### :gem: Type of change

### :scroll: Example code (TypeScript)
```ts
// before:
    function (this: Service, context: Context<any, any>, response: any): any {
      // Error: Property 'locals' does not exist on type 'Context<any, any>'.ts(2339)
      // context.locals.orderAction = "activate";
      (context as any).locals.orderAction = "activate";
      return response;
    },

// after:
    function (this: Service, context: Context<any, any>, response: any): any {
      // OK now
      context.locals.orderAction = "activate";
      return response;
    },
``` 

Not needed.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -527,6 +527,8 @@ declare namespace Moleculer {
 		needAck: boolean | null;
 		ackID: string | null;
 
+		locals: any | null;
+
 		level: number;
 
 		params: P;
--- a/index.d.ts
+++ b/index.d.ts
@@ -527,7 +527,7 @@ declare namespace Moleculer {
 		needAck: boolean | null;
 		ackID: string | null;
 
-		locals: any | null;
+		locals: GenericObject;
 
 		level: number;
 


index.d.ts
```diff
@@ -527,6 +527,8 @@ declare namespace Moleculer {
 		needAck: boolean | null;
 		ackID: string | null;
 
+		locals: any | null;
```
icebob, line 530: Please be more strict. `ctx.local` always an object and can't be null. So use `local: GenericObject`

icebob (changes requested):

index.d.ts
wiva-koikoi: Thank you.
Changed to
```ts
locals: GenericObject;
```

icebob (approved):
Thanks

Issue #623, opened 2019-11-26
Declare plugins as peerDependencies


oliversalzburg:
**Is your feature request related to a problem? Please describe.**
We use a package manager that utilizes symlinks to build the dependency tree from a flat storage area of all installed packages. Because of that, packages can not `require()` other packages that live closer to the root of the dependency tree, because their on-disk location doesn't have the required dependency as a parent.

**Describe the solution you'd like**
npm allows to declare `peerDependencies` to advertise the desire to a load a module that is not directly depended upon. Declaring a module as a peer dependency ensures that it will be made available to the depending module if it exists in the dependency tree.

Thus, declaring all dependencies moleculer _might_ want to require should be declared as `peerDependencies`. This will also ensure that it is clear which versions of the dependency moleculer is designed to run with.

**Describe alternatives you've considered**
Peering can be manually configured in our package manager and that would work just fine for us. I would consider adding the peer dependencies _more correct_ for the reasons given above.

**Additional context**
It seems some work in this area existed once: https://github.com/moleculerjs/moleculer/pull/37

icebob:
Could you explain it with an example?

oliversalzburg:
In the AMQP transporter, you [require `amqplib`](https://github.com/moleculerjs/moleculer/blob/master/src/transporters/amqp.js#L120). moleculer does not declare a dependency on `amqplib`, so this `require()` will fail, unless a dependent of `moleculer` installed a version of `amqplib` into the dependency tree structure into a place where it can be loaded by NodeJS from within the moleculer source code.

The same pattern happens for several other "pluggable" modules in the framework.

This is problematic, because you don't know which version of the required library you will receive and you can't make guarantees regarding compatbilitiy. Thus, the accepted version should be declared in a [peerDependencies](https://nodejs.org/es/blog/npm/peer-dependencies/) section in the `package.json`.

Additionally, there is really no guarantee that the software will be installed in a way that `moleculer` would be able to `require()` these modules without declaring them as peers. While the npm and yarn default layouts usually satisfy this requirement, the NodeJS module loader does not dictate this layout in any way.

The real only proper way to resolve this is to have the dependent code pass their instance of the module into your code. Then they can decide which library, with which modifications they get to load and to pass into moleculer. If you want more control in moleculer, declare the dependency in moleculer.

icebob:
But if I add all available pluggable lib dependencies as peerDependency, but you don't use them, NPM will warn you that these many peer dependency is not installed.

[...]

PR #828, opened 2020-10-17
Explicitly declare optional dependencies

oliversalzburg:
## :memo: Description
Declares all optional dependencies explicitly.

I hope I didn't miss any dependencies that should be on the list

### :dart: Relevant issues
#623

### :gem: Type of change

It hasn't been tested explicitly beyond running `npm install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,70 @@
     "node-fetch": "^2.6.1",
     "recursive-watch": "^1.1.4"
   },
+  "peerDependencies": {
+    "amqplib": "^0.6.0",
+    "avsc": "^5.4.22",
+    "bunyan": "^1.8.14",
+    "etcd3": "^1.0.2",
+    "ioredis": "^4.17.3",
+    "jaeger-client": "^3.18.1",
+    "kafka-node": "^5.0.0",
+    "mqtt": "^4.2.1",
+    "msgpack5": "^4.2.1",
+    "nats": "^1.4.12",
+    "node-nats-streaming": "^0.0.51",
+    "notepack.io": "^2.3.0",
+    "pino": "^6.6.1",
+    "protobufjs": "^6.10.1",
+    "redlock": "^4.1.0"
+  },
+  "peerDependenciesMeta": {
+    "amqplib": {
+      "optional": true
+    },
+    "avsc": {
+      "optional": true
+    },
+    "bunyan": {
+      "optional": true
+    },
+    "etcd3": {
+      "optional": true
+    },
+    "ioredis": {
+      "optional": true
+    },
+    "jaeger-client": {
+      "optional": true
+    },
+    "kafka-node": {
+      "optional": true
+    },
+    "mqtt": {
+      "optional": true
+    },
+    "msgpack5": {
+      "optional": true
+    },
+    "nats": {
+      "optional": true
+    },
+    "node-nats-streaming": {
+      "optional": true
+    },
+    "notepack.io": {
+      "optional": true
+    },
+    "pino": {
+      "optional": true
+    },
+    "protobufjs": {
+      "optional": true
+    },
+    "redlock": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": ">= 10.x.x"
   },


icebob:
My problem this new peerDependencies works (as expected) only with NPM 7. In previous versions, NPM will install all peer dependencies even they are not needed.

oliversalzburg:
To be a structurally coherent project, you should only load code that you depend upon. If you're safeguarding your imports with `if` statements **and** you require the matching change to be made to consumer dependencies, that's not a good design.

You have the `require` calls in your code. Your code depends on those modules. _Not_ installing them through `peerDependenciesMeta` is the hack. Not installing them, because you're offloading the installation of the correct, supported version to the consumer of your module *and* relying on ultimately broken hoisting behavior for your code to work is even worse and not even a hack.

Refusing to change this because users of outdated package managers will download packages that your code (optionally) depends on, is the wrong reason to reject this.

If you'd want these modules to be _truely_ optional, then their `require` statements would have to live in their own modules that are optional to install into moleculer, or consumers must pass in their own `require`'d modules (which then create possible version conflicts, because you're not declaring your expected dependencies properly).

shawnmcknight:
> My problem this new peerDependencies works (as expected) only with NPM 7. In previous versions, NPM will install all peer dependencies even they are not needed.

@icebob The behavior is _best_ in v7 in that it will warn you if the peer dependency has been installed but the semver range has not been met.  However, it should suppress warnings as of [6.11.10](https://github.com/npm/cli/blob/e09bdc142a4d9fe247927eb82a49720d5370b293/changelogs/CHANGELOG-6.md#features-2).  With the recent breaking change to NATS and without a really clear way to illustrate supported version ranges today, it might make some sense to push this through.  I also have an open branch for this same change which is probably a little more up to date if you agree that we should move forward with this.

icebob:
And how yarn handles it?

shawnmcknight:
> And how yarn handles it?

Looks like for 1.x it silences warnings as of [1.13.0](https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md#1130) and for 2.x it silences warnings as of [2.0.0](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md#package-manifests-packagejson).  1.13.0 was released 2018-12-18 and its always been in 2.x.

icebob:
@shawnmcknight could you up-to-date this PR, please?

package.json
```diff
@@ -120,6 +120,70 @@
     "node-fetch": "^2.6.1",
     "recursive-watch": "^1.1.4"
   },
+  "peerDependencies": {
```
shawnmcknight, line 123: I had recently done this analysis because I didn't realize that there was an open PR already out here.  Comparing this to what I had come up with there's a few differences:

1. I had found a few additional dependencies that fell into this category.
2. My approach was to avoid pinning to a specific patch release.  So for versions >= 1 I used `^{major}.0.0` and for versions < 1 I used `^0.{minor}.0`.

That led to a dependency list that looked like:

```json
  "peerDependencies": {
    "amqplib": "^0.7.0",
    "avsc": "^5.0.0",
    "bunyan": "^1.0.0",
    "cbor-x": "^0.6.0",
    "dd-trace": "^0.31.0",
    "debug": "^4.0.0",
    "etcd3": "^1.0.0",
    "ioredis": "^4.0.0",
    "jaeger-client": "^3.0.0",
    "kafka-node": "^5.0.0",
    "log4js": "^6.0.0",
    "mqtt": "^4.0.0",
    "msgpack5": "^5.0.0",
    "nats": "^1.0.0",
    "node-nats-streaming": "^0.3.0",
    "notepack.io": "2.0.0",
    "pino": "^6.0.0",
    "protobufjs": "^6.0.0",
    "redlock": "^4.0.0",
    "rhea-promise": "^1.0.0",
    "thrift": "^0.14.0",
    "winston": "^3.0.0"
  },
```

Could you update this list?

package.json
```diff
@@ -120,6 +120,70 @@
...
+    "msgpack5": "^4.2.1",
+    "nats": "^1.4.12",
+    "node-nats-streaming": "^0.0.51",
+    "notepack.io": "^2.3.0",
+    "pino": "^6.6.1",
+    "protobufjs": "^6.10.1",
+    "redlock": "^4.1.0"
+  },
+  "peerDependenciesMeta": {
+    "amqplib": {
```
shawnmcknight, line 141: This is a stylistic nitpick, but I felt that newlines on these braces adds a lot of noise since there's only one property over and over.  My list looked like:

```json
  "peerDependenciesMeta": {
    "amqplib": { "optional": true },
    "avsc": { "optional": true },
    "bunyan": { "optional": true },
    "cbor-x": { "optional": true },
    "dd-trace": { "optional": true },
    "debug": { "optional": true },
    "etcd3": { "optional": true },
    "ioredis": { "optional": true },
    "jaeger-client": { "optional": true },
    "kafka-node": { "optional": true },
    "log4js": { "optional": true },
    "mqtt": { "optional": true },
    "msgpack5": { "optional": true },
    "nats": { "optional": true },
    "node-nats-streaming": { "optional": true },
    "notepack.io": { "optional": true },
    "pino": { "optional": true },
    "protobufjs": { "optional": true },
    "rhea-promise": { "optional": true },
    "thrift": { "optional": true },
    "winston": { "optional": true }
  },
```

shawnmcknight (changes requested):
Thanks for updating this.  I have a few comments for feedback.

package.json
oliversalzburg: If you have a list of extensions+versions that you like and you have a different formatting that you like, you can just fix this yourself. I don't see the point in having me doing this extra work, if you've already done it.

shawnmcknight:
Superseded by #911.  Closing.

PR #835, opened 2020-11-03
escape for non-format asterisks

OutdatedVersion:
## :memo: Description

This change adds escapes to the protocol documentation. Before, the asterisks were picked up as formatting sequences causing visual display issues.

### :gem: Type of change

- [x] This change ~requires~ is a documentation update


icebob (approved):
Thanks

PR #838, opened 2020-12-08
Raising an info log event instead of warn on node expected disconnection

alexby:
## :memo: Description

Without this change we're raising a warn event in logger when any node is gracefully disconnected. But this event does not require any attention, so it should not be warning, but just an info level event.

### :gem: Type of change

- [x] "Test NodeCatalog.disconnected" -> "should call disconnected & unregister services": the `info` is called with a particular text, the `warn` is not called;
- [x] "Test NodeCatalog.disconnected" -> "should call disconnected & unregister services (unexpected)": the `info` is not called, the `warn` is called with a particular text;

### DIFF
--- a/src/registry/node-catalog.js
+++ b/src/registry/node-catalog.js
@@ -194,7 +194,10 @@ class NodeCatalog {
 
 			this.registry.updateMetrics();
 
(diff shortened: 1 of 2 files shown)


icebob (approved):
Great, thanks!

PR #843, opened 2020-12-14
Add availability marker to ServiceBroker#waitForServices API

OutdatedVersion:
## :memo: Description

Previously, both the return value and data included on the error thrown did not inform us which services we requested to watch became available or remained unavailable. Here we enhance the API to include that information so consumers can take informed next steps.

The selected API is what I believe to be the best way forwards _but_ it is breaking. Here is an example of a non-breaking alternative:
```js
try {
  await broker.waitForServices([
    { serviceName: 'posts', version: 2 },
    { serviceName: 'users', version: 1 },
  ]);
} catch (error) {
  // error.data.unavailable === ['v2.posts']
  // error.data.available === ['v1.users']
  logger.error(`Oh no, "${error.data.unavailable.join(', ')}" didn't make it`);				
}
``` 

### :dart: Relevant issues

None. I couldn't find any issue discussing anything similar either.

### :gem: Type of change

### :scroll: Example code
```js
try {
  await broker.waitForServices([
    { serviceName: 'posts', version: 2 },
    { serviceName: 'users', version: 1 },
  ]);
} catch (error) {
  const unavailableServices = error.data.statuses.filter(s => !s.available).map(s => s.name);
  logger.error(`Oh no, "${unavailableServices.join(', ')}" didn't make it`);				
}
``` 

- [x] Test A: Uh, not sure what to put here but I wrote unit tests? Hasn't been tested in an integration yet.
- [ ] Test B

   - I followed the style around the changed code.. let me know if something looks off. An auto-formatter would be helpful.

### DIFF
--- a/src/service-broker.js
+++ b/src/service-broker.js
@@ -959,26 +959,31 @@ class ServiceBroker {
 		})));
 
 		if (serviceNames.length == 0)
-			return this.Promise.resolve();
+			return this.Promise.resolve({ services: [], statuses: [] });
 
 		logger.info(`Waiting for service(s) '${serviceNames.join(", ")}'...`);
 
 		const startTime = Date.now();
 		return new this.Promise((resolve, reject) => {
 			const check = () => {
-				const count = serviceNames.filter(fullName => {
-					return this.registry.hasService(fullName);
+				const serviceStatuses = serviceNames.map(serviceName => {
+					return {
+						name: serviceName,
+						available: this.registry.hasService(serviceName),
+					}
 				});
 
-				if (count.length == serviceNames.length) {
+				const availableServiceCount = serviceStatuses.filter(s => s.available).length;
+
+				if (availableServiceCount == serviceNames.length) {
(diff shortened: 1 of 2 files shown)


icebob (commented):
Good job! Yeah the current implementation causes a breaking change in the error handling. If somebody processed the `err.services` as string array, now they won't able to do.
To avoid it, we should stay the `err.services` and add the new response under another property name, like `statuses` in both response. And in this case, I can merge it even into 0.14. WDYT?

OutdatedVersion:
> Good job! Yeah the current implementation causes a breaking change in the error handling. If somebody processed the `err.services` as string array, now they won't able to do.
> To avoid it, we should stay the `err.services` and add the new response under another property name, like `statuses` in both response. And in this case, I can merge it even into 0.14. WDYT?

I've updated the implementation to be non-breaking; going with your `statuses` suggestion.

Another change here to note: when you pass an empty array of services this will now return an object like `{ services: [], statuses: [] }` instead of `undefined`.

icebob (approved):
Great job, thanks!

Issue #849, opened 2020-12-28
Plaintext TCP Transport nodes crash nodes with Encryption middleware


cintolas:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [ x] I am running the latest version
- [x ] I checked the documentation and found no answer
- [x ] I checked to make sure that this issue has not already been filed
- [ x] I'm reporting the issue to the correct repository

## Current Behavior
If you configure 2 services, one using Encryption and the other using plaintext, the encryption node will crash when it gets a message from the plaintext node

## Expected Behavior
Encryption node should not crash.  Either issue a warning or silently fail

## Failure Information
The decrypt function is not properly caught so it crashes

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Create a basic TCP transport node
2. Create a TCP transport node with the Middlewares.Transport.Encryption middleware
3. Launch both nodes
4. Wait for the communication to occur

### Reproduce code snippet
The following code snippet reproduces the issue.  Run one as is, and the other with the ENCRYPT environment variable set

```
const { ServiceBroker, Middlewares } = require('moleculer');
const crypto = require('crypto');


if(process.env.ENCRYPT) {
    const initVector = crypto.randomBytes(16);
    let str = crypto.createHash('sha256')
    .update(String("password"))
    .digest('base64')
    .substr(0, 32);

    let encryptConfig = {
        namespace: 'encrypt-example',
        nodeID: 'encrypt-me',
        transporter: {
            type: 'TCP'
        },
        middlewares: [Middlewares.Transmit.Encryption(str, "aes-256-cbc", initVector)]
    };
    let encryptBroker = new ServiceBroker(encryptConfig);
    encryptBroker.start();
    return;
}

let plainConfig = {
    namespace: 'encrypt-example',
    nodeID: 'plaintext',
    transporter: {
        type: 'TCP'
    },
};
let plainBroker = new ServiceBroker(plainConfig);
plainBroker.start();
return;
```
### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.14
* NodeJS version:  13.3
* Operating System: CentOS 7.9

### Failure Logs
```
internal/crypto/cipher.js:174
  const ret = this[kHandle].final();
                            ^

Error: error:0606506D:digital envelope routines:EVP_DecryptFinal_ex:wrong final block length
[...]

PR #850, opened 2020-12-28
Issue #849.  Catch decryption failure

cintolas:
If decryption fails, catch the error and silently continue.
This prevents a misconfigured node from crashing nodes using Middlewares.Transport.Encryption

## :memo: Description
A misconfigured node should not crash other nodes.
I wrapped the decrypt code in a try/catch block.  This will log the following error on the sender's side
[2020-12-28T22:09:47.008Z] WARN  plaintext/TRANSPORTER: Invalid incoming GOSSIP_REQ packet

### :dart: Relevant issues

#849 

### :gem: Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

### :scroll: Example code
```js

``` 

Tested against the breaking code in case 849

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code


icebob (changes requested):
Nice job, but please don't use silent catch, because it makes the debugging difficult. Please add a logger.error because it's not a normal case and the developer should know about it.

PR #854, opened 2021-01-14
Add default Context Meta properties

HazemKhaled:
### :gem: Type of change

### :scroll: Example code
```js

``` 

- [ ] Test A
- [ ] Test B

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -479,6 +479,15 @@ declare namespace Moleculer {
 
 	type ServiceActionsSchema = { [key: string]: ActionSchema | ActionHandler | boolean; };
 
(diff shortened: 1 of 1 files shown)


icebob:
I don't feel it's the good solution because these properties are relevant if the action handler via `moleculer-web`. What do you think @shawnmcknight ?

shawnmcknight:
Yes I would tend to agree. Since `moleculer-web` is an add-on, I don't feel like the types necessarily belong in `moleculer` proper. I wouldn't be opposed to that type being added to `moleculer-web` as a type alias that could be imported as a helper though. A user who wanted that type could replace the generic in `Context` with it or augment the `Context` interface if they wanted it throughout their environment.

PR #855, opened 2021-01-14
Fix version string build when an integer value is passed in string form

slinkardbrandon:
## :memo: Description

I got tripped up on something (that was my fault) when using the `broker.waitForServices()` method and passing an array of objects. In my case, serviceVersion was a string even though I thought it was a number. Below is an example as to how you can see how that might have occurred. I'm just hoping this change might prevent the pain for anyone else in the future.

```ts
    await broker.waitForServices([
      { name: serviceName, version: serviceVersion },
    ]);
```

### :dart: Relevant issues

N/A

### :gem: Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### :scroll: Example code
```js
  /**
   * This will now await for the following services to become available:
   * v1.foo & v2.bar
   * 
   * Previously this would wait for
   * v1.foo & 2.bar
   */ 
    await broker.waitForServices([
      { name: 'foo', version: 1 },
      { name: 'bar', version: '2` }
    ]);
``` 

- [x] Test A - I added to the version string building static method tests with a case that demonstrated the problem, then made it work

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -772,7 +772,7 @@ class Service {
 	 */
 	static getVersionedFullName(name, version) {
 		if (version != null) {
-			if (typeof(version) === 'number' || !isNaN(parseInt(version))) {
+			if (typeof(version) === 'number' || !isNaN(Number(version))) {
 				return `v${version}.${name}`;
 			}
 			return `${version}.${name}`;
--- a/src/service.js
+++ b/src/service.js
@@ -771,8 +771,12 @@ class Service {
 	 * @param {String|Number?} version
 	 */
 	static getVersionedFullName(name, version) {
-		if (version != null)
-			return (typeof(version) == "number" ? "v" + version : version) + "." + name;
+		if (version != null) {
+			if (typeof(version) === 'number' || !isNaN(parseInt(version))) {
+				return `v${version}.${name}`;
+			}
+			return `${version}.${name}`;
+		}
 
 		return name;
 	}
--- a/test/unit/service.spec.js
+++ b/test/unit/service.spec.js
@@ -1925,6 +1925,7 @@ describe("Test Service class", () => {
 			expect(Service.getVersionedFullName("posts")).toBe("posts");
 			expect(Service.getVersionedFullName("posts", 5)).toBe("v5.posts");
 			expect(Service.getVersionedFullName("posts", "testing")).toBe("testing.posts");
+			expect(Service.getVersionedFullName("posts", '5')).toBe("v5.posts");
 		});
 	});
 
--- a/test/unit/service.spec.js
+++ b/test/unit/service.spec.js
@@ -1926,6 +1926,7 @@ describe("Test Service class", () => {
 			expect(Service.getVersionedFullName("posts", 5)).toBe("v5.posts");
 			expect(Service.getVersionedFullName("posts", "testing")).toBe("testing.posts");
 			expect(Service.getVersionedFullName("posts", '5')).toBe("v5.posts");
+			expect(Service.getVersionedFullName("posts", '123abc')).toBe("123abc.posts");
 		});
 	});
 


src/service.js
```diff
@@ -771,8 +771,12 @@ class Service {
 	 * @param {String|Number?} version
 	 */
 	static getVersionedFullName(name, version) {
-		if (version != null)
-			return (typeof(version) == "number" ? "v" + version : version) + "." + name;
+		if (version != null) {
+			if (typeof(version) === 'number' || !isNaN(parseInt(version))) {
```
ngraef, line 775: This is a breaking change if `version = "123abc"`.

slinkardbrandon: Good call, I just updated this to use the Number constructor instead and I added a test case to validate that it works the way we would expect. 

Thanks!

ngraef: It's technically also breaking for anyone who is using `version = "5"` intentionally... I understand your pain here though.

slinkardbrandon: That's also true, I would think this would be a "good" breaking change from my perspective. Let me update the PR to flag that this is breaking and let @icebob think on it 🙂

icebob:
@slinkardbrandon thanks for the PR, but I share the viewpoint of @ngraef. Every change in this logic can cause a breaking change because users can use version as `5` or `"5"` but they are different in the current and all previous versions. 
So now I won't merge it but leave open.

slinkardbrandon:
@icebob and @ngraef what are your thoughts on simply logging a warning statement if a version like "5" is used? That way we can preserve the existing functionality but potentially warn developers of their goof with an entry to their logger.

0x0a0d:
> @icebob and @ngraef what are your thoughts on simply logging a warning statement if a version like "5" is used? That way we can preserve the existing functionality but potentially warn developers of their goof with an entry to their logger.

you can do this by using `patch-package` :D

test/unit/service.spec.js
```diff
@@ -1925,6 +1925,8 @@ describe("Test Service class", () => {
 			expect(Service.getVersionedFullName("posts")).toBe("posts");
 			expect(Service.getVersionedFullName("posts", 5)).toBe("v5.posts");
 			expect(Service.getVersionedFullName("posts", "testing")).toBe("testing.posts");
+			expect(Service.getVersionedFullName("posts", '5')).toBe("v5.posts");
```
icebob, line 1928: Codacy has a fix for the issue: [Strings must use doublequote.](https://app.codacy.com/gh/moleculerjs/moleculer/pullRequest?prid=8141838)
```suggestion
			expect(Service.getVersionedFullName("posts", "5")).toBe("v5.posts");
```

test/unit/service.spec.js
```diff
@@ -1925,6 +1925,8 @@ describe("Test Service class", () => {
...
+			expect(Service.getVersionedFullName("posts", '123abc')).toBe("123abc.posts");
```
icebob, line 1929: Codacy has a fix for the issue: [Strings must use doublequote.](https://app.codacy.com/gh/moleculerjs/moleculer/pullRequest?prid=8141838)
```suggestion
			expect(Service.getVersionedFullName("posts", "123abc")).toBe("123abc.posts");
```

src/service.js
```diff
@@ -771,8 +771,12 @@ class Service {
 	 * @param {String|Number?} version
 	 */
 	static getVersionedFullName(name, version) {
-		if (version != null)
-			return (typeof(version) == "number" ? "v" + version : version) + "." + name;
+		if (version != null) {
+			if (typeof(version) === 'number' || !isNaN(Number(version))) {
```
icebob, line 775: Codacy has a fix for the issue: [Strings must use doublequote.](https://app.codacy.com/gh/moleculerjs/moleculer/pullRequest?prid=8141838)
```suggestion
			if (typeof(version) === "number" || !isNaN(Number(version))) {
```

Issue #875, opened 2021-02-22
Circular dependency loop possible hot-reload middleware


nurdism:
- [x] I am running the latest version (0.14.12)
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Current Behavior
Circular dependency loop is possible in `processModule` function  in the `hot-reload` middleware, causing a loop in the runner command. I did some poking around and logged the file names it was importing and it got stuck in some sort of circular dependency loop:

```
/workspace/dist/entity/job/JobImport.js
/workspace/dist/lib/orm.js
/workspace/dist/util/entity.js
/workspace/dist/lib/errors.js
/workspace/dist/lib/errors/http.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/code.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/lib/errors/api.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/lib/errors/http.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/code.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/util/date.js
/workspace/dist/lib/orm.js
/workspace/dist/util/entity.js
/workspace/dist/lib/errors.js
/workspace/dist/lib/errors/http.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/code.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/lib/errors/api.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/lib/errors/http.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/code.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/lib/http/status.js
/workspace/dist/lib/http/status/message.js
/workspace/dist/util/date.js
... repeats
```

### Steps to Reproduce
I can reproduce this on my code base, but I can't figure out how it has become circular. I can not share my code base as its is closed source. I am using typeorm and I have a *ton* of potential circular dependencies. 

* Moleculer version: 0.14.12
* NodeJS version: 14x
[...]

PR #876, opened 2021-02-22
Potential fix for circular dependency loop #875

nurdism:
### :dart: Relevant issues
#875

### :gem: Type of change

Bug fix (non-breaking change which fixes an issue)

I have tested in my local environment with serval services, does not affect the functionality of hot-reload middleware.

- [ ] ~**I have added tests that prove my fix is effective or that my feature works**~
- [ ] ~**New and existing unit tests pass locally with my changes**~


nurdism:
Note, I only say "potential fix" because I'm not entirely sure what the `processModule` function really does, any not entirely sure if checking if a file has been processed is across the whole tree will introduce any new bugs. It *seems* to work just fine with the changes I've made.

icebob:
We can't merge without a reproducible example.

PR #907, opened 2021-04-10
Add optional peer dependencies

### LINKED ISSUE #623: Declare plugins as peerDependencies

shawnmcknight:
## :memo: Description

It is difficult to know what versions of optional peer dependencies are supported by moleculer like the pluggable transporters and serializers.  Previously, there was no appropriate location to store these in `package.json`.  `peerDependencies` would issue warnings (which may be invalid).  `optionalDependencies` would install them automatically whether you want them or not.  A (relatively) new feature has been added to npm/yarn which will respect a `peerDependenciesMeta` property in `package.json`.  This allows peer dependencies to be specified but also declared as optional per the [docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta).  This will provide a user of moleculer to have a better understanding of the supported ranges of the pluggable modules without receiving warnings about them not being installed.

This PR adds all of the optional logger, transport, serializer, caching, and tracing dependencies to `package.json` in this format.  I do not necessarily have the knowledge to determine what specific version ranges of these modules are truly supported, so I took my best guess.  I used the latest caret range major version of every dependency with the exception of `nats` which I know is not supported as of 2.0.0.  For every dependency with a version above 1.0.0, I used a format of `^{major}.0.0` and for every dependency below 1.0.0, I used a format of `^0.{minor}.0`.

### :dart: Relevant issues

Resolves #623.

### :gem: Type of change

I created a new shell repo and installed moleculer from my forked github branch.  I did not receive any peer dependency warnings.


shawnmcknight:
I just realized I foolishly missed another open PR for this already #828.  Closing this one out.

PR #912, opened 2021-04-19
Resolves #909 - Add Utils typings

### LINKED ISSUE #909: Utils type definitions not defined, preventing typescript import
## Current Behavior

`import { Utils } from 'moleculer';`
`Module '"moleculer"' has no exported member 'Utils'.ts(2305)`

## Expected Behavior

`import { Utils } from 'moleculer';`
`Utils.function()` * success

## Failure Information

`Utils` type declarations seem to not be defined, falling import using typescript. However, the module is exported as follows.
https://github.com/moleculerjs/moleculer/blob/c1aa5833c9701dbdec7516e3ee88d170dbef4162/index.js#L42

### Steps to Reproduce

When typescript:
`import { Utils } from 'moleculer';`

### Context

* Moleculer version: v0.14.13
* NodeJS version: latest
* Operating System: any

srMarquinho:
## :memo: Description

Utils functions in typescript are not possible to import.
Needed to use one of the functions in our typescript project.

### :dart: Relevant issues
#909 

### :gem: Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)

### :scroll: Example code

```ts
import { Utils } from 'moleculer';

Utils.match('this.is.a.text',  '**');
``` 

This has been tested on a local typescript project. According to the example given above.

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] **New and existing unit tests pass locally with my changes**
- [ x] I have commented my code, particularly in hard-to-understand areas

NOTE: To follow the file standards, no description comments added.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1590,7 +1590,7 @@ declare namespace Moleculer {
 		function isString(str: unknown): str is string;
 		function isObject(obj: unknown): obj is object;
 		function isPlainObject(obj: unknown): obj is object;
-		function isDate(date: unknown): data is Date;
+		function isDate(date: unknown): date is Date;
 		function flatten<T>(arr: readonly T[] | readonly T[][]): T[];
 		function humanize(millis?: number | null): string;
 		function generateToken(): string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1586,7 +1586,7 @@ declare namespace Moleculer {
 	}
 
 	namespace Utils {
-		function isFunction(func: unknown): func is function;
+		function isFunction(func: unknown): boolean;
 		function isString(str: unknown): str is string;
 		function isObject(obj: unknown): obj is object;
 		function isPlainObject(obj: unknown): obj is object;
@@ -1597,8 +1597,8 @@ declare namespace Moleculer {
 		function removeFromArray<T>(arr: T[], item: T): T[];
 		function getNodeID(): string;
 		function getIpList(): string[];
-		function isPromise(promise: unknown): promise is Promise;
-		function polyfillPromise(P: Promise): void;
+		function isPromise<T>(promise: unknown): promise is Promise<T>;
+		function polyfillPromise<T>(P: Promise<T>): void;
 		function clearRequireCache(filename: string): void;
 		function match(text: string, pattern: string): boolean;
 		function deprecate(prop: unknown, msg?: string): void;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1586,7 +1586,7 @@ declare namespace Moleculer {
 	}
 
 	namespace Utils {
-		function isFunction(func: unknown): boolean;
+		function isFunction(func: unknown): func is Function;
 		function isString(str: unknown): str is string;
 		function isObject(obj: unknown): obj is object;
 		function isPlainObject(obj: unknown): obj is object;
@@ -1598,7 +1598,7 @@ declare namespace Moleculer {
 		function getNodeID(): string;
 		function getIpList(): string[];
 		function isPromise<T>(promise: unknown): promise is Promise<T>;
-		function polyfillPromise<T>(P: Promise<T>): void;
+		function polyfillPromise(P: typeof Promise): void;
 		function clearRequireCache(filename: string): void;
 		function match(text: string, pattern: string): boolean;
 		function deprecate(prop: unknown, msg?: string): void;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1586,24 +1586,24 @@ declare namespace Moleculer {
 	}
 
 	namespace Utils {
-		function isFunction(func: any): boolean;
-		function isString(str: any): boolean;
-		function isObject(obj: any): boolean;
-		function isPlainObject(obj: any): boolean;
-		function isDate(date: any): boolean;
-		function flatten(arr: any[]): any[];
-		function humanize(millis: number): string;
+		function isFunction(func: unknown): func is function;
+		function isString(str: unknown): str is string;
+		function isObject(obj: unknown): obj is object;
+		function isPlainObject(obj: unknown): obj is object;
+		function isDate(date: unknown): data is Date;
+		function flatten<T>(arr: readonly T[] | readonly T[][]): T[];
+		function humanize(millis?: number | null): string;
 		function generateToken(): string;
-		function removeFromArray(arr: any[], item: any): any[];
+		function removeFromArray<T>(arr: T[], item: T): T[];
 		function getNodeID(): string;
-		function getIpList(): any[];
-		function isPromise(promise: any): boolean;
-		// function polyfillPromise(P: Promise): any; // From Utils -> NOT USED & NOT TESTED YET !!!
+		function getIpList(): string[];
+		function isPromise(promise: unknown): promise is Promise;
+		function polyfillPromise(P: Promise): void;
 		function clearRequireCache(filename: string): void;
 		function match(text: string, pattern: string): boolean;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1584,6 +1584,29 @@ declare namespace Moleculer {
 			delimiter(value: string): void;
 		}
 	}
+
+	namespace Utils {
+		function isFunction(func: any): boolean;
```
shawnmcknight, line 1589: The "is" series of functions are basically acting as typeguards, and we can express that in the syntax here.  For instance, `isString(str: unknown): str is string;`

I also think its best to use `unknown` instead of `any` on these as it more appropriately describes the expected inputs.

index.d.ts
```diff
@@ -1584,6 +1584,29 @@ declare namespace Moleculer {
...
+		function removeFromArray(arr: any[], item: any): any[];
+		function getNodeID(): string;
+		function getIpList(): any[];
+		function isPromise(promise: any): boolean;
+		// function polyfillPromise(P: Promise): any; // From Utils -> NOT USED & NOT TESTED YET !!!
+		function clearRequireCache(filename: string): void;
+		function match(text: string, pattern: string): boolean;
+		function deprecate(prop: any, msg: string): void;
+		function safetyObject(obj: any, options: any): any;
+		function dotSet(obj: any, path: string, value: any): any;
```
shawnmcknight, line 1606: I'm cribbing a bit from lodash's types for `set` here, but its effectively the same concept.  lodash has this defined as:

`set<T extends object>(object: T, path: PropertyPath, value: any): T;`

To try to keep this somewhat simple, I think `dotSet<T extends object>(obj: T, path: string, value: unknown): T;` makes the most sense.

index.d.ts
```diff
@@ -1584,6 +1584,29 @@ declare namespace Moleculer {
...
+		function humanize(millis: number): string;
+		function generateToken(): string;
+		function removeFromArray(arr: any[], item: any): any[];
+		function getNodeID(): string;
+		function getIpList(): any[];
+		function isPromise(promise: any): boolean;
+		// function polyfillPromise(P: Promise): any; // From Utils -> NOT USED & NOT TESTED YET !!!
+		function clearRequireCache(filename: string): void;
+		function match(text: string, pattern: string): boolean;
+		function deprecate(prop: any, msg: string): void;
```
shawnmcknight, line 1604: @icebob Can you fill me in on what this function is used for?  I don't see it referenced anywhere nor is the `deprecateList` it mutates used anywhere.  I'm not sure what the inputs to this function should be, although the jsdoc indicates it as `Object|Function|string`.  Is there any real expectation of what the input to this should be?

shawnmcknight (changes requested):
Thank you for taking care of this.  I had a few notes regarding some tightening up some of the any's a bit.

index.d.ts
srMarquinho: Thank you @shawnmcknight for your review. I'm then waiting from @icebob on this to update.

icebob: Yeah, the input can be anything. You can use it to write a deprecate warning in the modules. E.g if there is a deprecated method in a class, and somebody call, that you can call it with the method. So this function will store the pointer of method and won't print the deprecate messages all times, just only one. But if there is a deprecated broker option, in this case, you can call it with a string. So normally it will be called with (Object,Function, String), but can be anything. 

Currently it's not used because there is no deprecation in the codebase, but it would be used in the future.

index.d.ts
srMarquinho: I'm unfamiliar with this syntax. I have updated the PR. Please let me know if this is correct. Thank you!

index.d.ts
```diff
@@ -1584,6 +1584,29 @@ declare namespace Moleculer {
...
+		function isPlainObject(obj: any): boolean;
+		function isDate(date: any): boolean;
+		function flatten(arr: any[]): any[];
+		function humanize(millis: number): string;
+		function generateToken(): string;
+		function removeFromArray(arr: any[], item: any): any[];
+		function getNodeID(): string;
+		function getIpList(): any[];
+		function isPromise(promise: any): boolean;
+		// function polyfillPromise(P: Promise): any; // From Utils -> NOT USED & NOT TESTED YET !!!
```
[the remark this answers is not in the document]
srMarquinho: `polyfillPromise(P: Promise): void;` seems to be invalid. I've added as `polyfillPromise<T>(P: Promise<T>): void;` not sure this is correct either.

src/utils.js
```diff
@@ -173,7 +173,6 @@ const utils = {
 	/**
 	 * Polyfill a Promise library with missing Bluebird features.
 	 *
-	 * NOT USED & NOT TESTED YET !!!
```
srMarquinho, line 176: Removed this comment as I believe there is a test for it.

icebob: Thanks, I missed to remove it after I've used in the broker.

index.d.ts
```diff
@@ -1584,6 +1584,29 @@ declare namespace Moleculer {
...
+		function isPlainObject(obj: unknown): obj is object;
+		function isDate(date: unknown): date is Date;
+		function flatten<T>(arr: readonly T[] | readonly T[][]): T[];
+		function humanize(millis?: number | null): string;
+		function generateToken(): string;
+		function removeFromArray<T>(arr: T[], item: T): T[];
+		function getNodeID(): string;
+		function getIpList(): string[];
+		function isPromise<T>(promise: unknown): promise is Promise<T>;
+		function polyfillPromise<T>(P: Promise<T>): void;
```
shawnmcknight, line 1601: After seeing this one, I realize I steered you wrong.  I think this should be:
`function polyfillPromise(p: typeof Promise): void;`

index.d.ts
```diff
@@ -1584,6 +1584,29 @@ declare namespace Moleculer {
 			delimiter(value: string): void;
 		}
 	}
+
+	namespace Utils {
+		function isFunction(func: unknown): boolean;
```
shawnmcknight, line 1589: Return type here can be `func is Function` to match the other typeguard signatures.

shawnmcknight (commented):
This looks great.  I had one note on one last typeguard-style signature and the `polyFillPromise` where I gave you some incorrect guidance.

index.d.ts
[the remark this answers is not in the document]
shawnmcknight: Ahh I had missed your reply here.  Yes, my suggestion was wrong.  I've noted another comment with what should be correct.

index.d.ts
shawnmcknight: Sorry, I should have looked at your replies here before I did a follow up review.  You did great with the typeguard syntax.  If you're interested, here are some notes on it: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-6.html#user-defined-type-guard-functions.  (Side-note, the typeguard syntax also works for things like array filters to narrow the result).

The function one I noted in my follow-up review.

index.d.ts
srMarquinho: Ah ha. That's the one. Thanks.

index.d.ts
srMarquinho: Thank you for this reference. 👍

srMarquinho:
Not sure that failed action is related. Also, I can't seem to be able to re-run it.

icebob:
I've re-run the CI, now it's green.

shawnmcknight (approved):
Looks great! Thank you!

icebob:
Thanks @srMarquinho and @shawnmcknight !

PR #918, opened 2021-04-30
Fix: etcd3-discoverer did not recover from lease-loss

SockeThe2nd:
## :memo: Loosing the etcd3 lease for heartbeat or info was not recovered

In the current implementation, the etcd3-discoverer does not recover if the lease for heartbeat or info is lost. This can happen when the node is experiencing high cpu load. The bug is reproducable by simply keeping the process locked by using a debugger until the lease is lost.

I implemented an handler for the lease-lost event as described [here](https://microsoft.github.io/etcd3/classes/lease.html)

The handler closes and removes the current lease. If the node is connected, I immediatly try to send a new heartbeat or info. Since the lease is removed, this will generate a new lease.

### :dart: Relevant issues

I did not open an issue, but described my problem on the [discord] (https://discord.com/channels/585148559155003392/585149564135145493/837299198629117974)

### :gem: Type of change

I updated the existing test for the etcd3-discoverer to check if the event is bound correctly. All tests pass locally.

I tested the implementation locally by starting two services on seperate nodes and debugged one of them until the lease on the etcd-keys expired. I then resumed normal execution. the service logged the error and recovered as expected.

  - I did extend the existing tests to check if the handler is bound, but did not write a test to check the failover itself. Help is welcome.


icebob:
Nice, I will check it soon.

icebob (approved):

PR #919, opened 2021-05-03
Promise all settled

intech:
## :memo: Description

Add options `settled` for `mcall`

### DIFF
--- a/test/unit/utils.spec.js
+++ b/test/unit/utils.spec.js
@@ -733,21 +733,23 @@ describe("Test utils.polyfillPromise", () => {
 		});
 	});
 
-	describe("Test Promise.promiseAllControl", () => {
-		it("should be rejected", () => {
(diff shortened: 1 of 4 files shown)


test/unit/utils.spec.js
```diff
@@ -733,4 +733,23 @@ describe("Test utils.polyfillPromise", () => {
...
+						{ status: "fulfilled", value: "First" },
+						{ status: "fulfilled", value: "Second" },
+						{ status: "fulfilled", value: "Third" },
+						{ status: "rejected", reason: "Error" }
+					]);
+				});
+			});
+		});
+	}
 });
```
icebob, line 755: please add another test where the `settled` is `false` to cover both case.

icebob (changes requested):

icebob (approved):
Awesome, thanks!

Issue #925, opened 2021-05-10
In the custom validation method has access to the broker or context


intech:
We need to think about how to think about how we integrate the fastest-validator with the moleculer so that we can use custom validation in the context of the moleculer.

```js
{
    type: "string", max: 4, min: 3, custom: (...args) => {
        // Access to the broker or context
        return true;
    }
}
```

PR #926, opened 2021-05-10
feat(validator): add meta in check

intech:
Related issue #925


test/unit/validators/base.spec.js
```diff
@@ -90,7 +90,7 @@ describe("Test BaseValidator 'middleware' method", () => {
 			// Call wrapped function
 			return wrapped(ctx).then(() => {
 				expect(checkGood).toHaveBeenCalledTimes(1);
-				expect(checkGood).toHaveBeenCalledWith({ "id": 5, "name": "John" });
```
icebob, line 93: Please leave the full object def `{ "id": 5, "name": "John" }`. It's a kind of protection checking the params whether is manipulated. The `ctx.params` hide it.

src/validators/base.js
```diff
@@ -67,7 +67,7 @@ class BaseValidator {
 				if (action[paramName] && typeof action[paramName] === "object") {
 					const check = self.compile(action[paramName]);
 					return function validateContextParams(ctx) {
-						let res = check(ctx.params != null ? ctx.params : {});
+						let res = check(ctx.params != null ? ctx.params : {}, { meta: ctx });
```
icebob, line 70: I'm thinking on that it's enough that we use the whole `meta` to point to the context. Because if the future we want to add more "things" to the meta, we can't, or it can cause breaking change. So maybe it can `{meta: { ctx } }`.

test/unit/validators/fastest.spec.js
```diff
@@ -84,3 +84,53 @@ describe("Test Validator.convertSchemaToMoleculer", () => {
...
+	it("should validate with meta", () => {
+		let p = { c: "asd" };
+		return broker.start().then(() => broker.call("test.withCustomValidation", p, { meta: { isTest: true } }))
+			.then(res => {
+				expect(res).toEqual(123);
+			});
+	});
+
+	it("should throw ValidationError without meta", () => {
+		let p = { c: "asd" };
```
icebob, line 129: If you add assertion into the `catch` case, you should add `expect.assertions(2)`. Because the catch is not called Jest will report it as success. But this command tells the Jest that two assertions should run.

icebob (changes requested):
Thanks, I wrote some comments.
And I think we can make the test cases simpler. I will add it in a commit to the PR.

src/validators/base.js
intech: Are we sure we want to do so much nesting of objects with 1 property?
If we want to add something else in the future, then we can do:
```js
{ meta: { ...ctx, ...etc }}
```

icebob: ok.

icebob (approved):

PR #945, opened 2021-06-18
nats v2 unit tests

AndreMaz:
- I've wrapped v1 tests in a dedicated `describe()` but didn't lint the code. I've done this to be easier to track the changes. After merging you should lint and reformat the code.
- In nats docs it says that `If the connection is dropped, the client will attempt to reconnect.` so I didn't add a test for it


icebob (approved):
Thanks, good job! You can merge it into `nats2` branch.

Issue #949, opened 2021-06-23
Typescript: ThisType in ActionHandler is not work


moonrailgun:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Current Behavior

For example code below:
```typescript
    this.parseServiceSchema({
      name: 'user',
      mixins: [DbService],
      actions: {
        login: {
          rest: 'POST /login',
          async handler(ctx: Context) {
            return {
              id: 1,
              username: 'test',
            };
          },
        },
      },
    });
```

I try to typing `this.` and not any type suggest display.

When i try to check the type of `this`, just any.

## Expected Behavior

Maybe we should change 
```typescript
type ActionHandler<T = any> = ((ctx: Context<any, any>) => Promise<T> | T) & ThisType<Service>;
```
to
```typescript
type ActionHandler<T = any> = ((this: Service, ctx: Context<any, any>) => Promise<T> | T);
```

`ThisType` looks like use in parent, not in `ActionHandler` self

https://github.com/moleculerjs/moleculer/blob/7b94f7cac722f5f82a1b5ed8906a915fa586b79f/index.d.ts#L47

shawnmcknight:
It appears that `ThisType` is only to be used with objects and not with methods, so the typing does appear to be incorrect.  Your proposal looks correct to me -- can you create a PR?

moonrailgun:
@shawnmcknight .Hi, thanks for your reply. I submmit a `pr` for my mind, pls take a look

dprslt:
Hi, 

I'm not sure if this is related, but i cannot use this with typescript in the following snippet : 

```ts
   public constructor(public broker: ServiceBroker) {
        super(broker);
        this.parseServiceSchema({
            name: "greeter",
            actions:{
                hello: {
                    async handler(): Promise<string> {
                        return this.ActionHello();
                    },
                },
                welcome: {
                    
                    params: {
                        name: "string",
                    },
                    async handler(ctx: Context<{name: string}>): Promise<string> {
                        await this.actions.hello(null, {parentCtx: ctx})
                        return this.ActionWelcome(ctx.params.name);
                    },
                },
            },
        });
    }
...
}
```

This file does not compile with the following error from ts : 

```
[...]

PR #950, opened 2021-06-25
fix definition file ThisType Usage

moonrailgun:
## :memo: Description

I think the hardcode in typescript which define
`this` type in function arguments.

It's more intuitive to user.

Reference: https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function

All changes is pass in my local repo. except
`ServiceEventHandler | ServiceEventLegacyHandler`
Becase of typescipt cannot make infer between two function

### :dart: Relevant issues

#949 

### :gem: Type of change

### :scroll: Example code
```js

``` 

- [x] Test A
```typescript
this.parseServiceSchema({
  methods: {
    foo(bar) {
      this; // check this
    }
  },
})
```
- [x] Test B
```typescript
this.parseServiceSchema({
  methods: {
    actions(ctx) {
      this; // check this
    }
  },
})
```
- [x] Test C
```typescript
broker.middlewares.add(function () {
  this; // check this
  return {};
});
```
- [x] Test D
```typescript
broker.services[0].actions = {
  foo(params, opts): any {
    this; // check this
  },
};
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,7 +44,10 @@ declare namespace Moleculer {
 		trace(...args: any[]): void;
 	}
 
-	type ActionHandler<T = any> = ((ctx: Context<any, any>) => Promise<T> | T) & ThisType<Service>;
+	type ActionHandler<T = any, S = ServiceSettingSchema> = (
+		this: Service<S>,
+		ctx: Context<any, any>
+	) => Promise<T> | T;
 	type ActionParamSchema = { [key: string]: any };
 	type ActionParamTypes =
 		| "any"
@@ -595,9 +598,15 @@ declare namespace Moleculer {
 		[name: string]: any;
 	}
 
-	type ServiceEventLegacyHandler = ((payload: any, sender: string, eventName: string, ctx: Context) => void) & ThisType<Service>;
+	type ServiceEventLegacyHandler = (
+		this: Service,
+		payload: any,
+		sender: string,
+		eventName: string,
+		ctx: Context
+	) => void;
 
-	type ServiceEventHandler = ((ctx: Context) => void) & ThisType<Service>;
+	type ServiceEventHandler = (this: Service, ctx: Context) => void;
 
 	interface ServiceEvent {
 		name?: string;
@@ -611,7 +620,12 @@ declare namespace Moleculer {
 
 	type ServiceEvents = { [key: string]: ServiceEventHandler | ServiceEventLegacyHandler | ServiceEvent };
 
-	type ServiceMethods = { [key: string]: ((...args: any[]) => any) } & ThisType<Service>;
+	type ServiceMethods = {
+		[key: string]: <S = ServiceSettingSchema>(
+			this: Service,
+			...args: any[]
+		) => any;
+	};
 
 	type CallMiddlewareHandler = (actionName: string, params: any, opts: CallingOptions) => Promise<any>;
 	type Middleware = {
@@ -624,7 +638,10 @@ declare namespace Moleculer {
 			| ((handler: CallMiddlewareHandler) => CallMiddlewareHandler)
 	}
 
-	type MiddlewareInit = (broker: ServiceBroker) => Middleware & ThisType<ServiceBroker>;
+	type MiddlewareInit = (
+		this: ServiceBroker,
+		broker: ServiceBroker
+	) => Middleware;
 	interface MiddlewareCallHandlerOptions {
 		reverse?: boolean
 	}
@@ -682,7 +699,15 @@ declare namespace Moleculer {
 		[name: string]: any;
 	}
 
-	type ServiceAction = (<T = Promise<any>, P extends GenericObject = GenericObject>(params?: P, opts?: CallingOptions) => T) & ThisType<Service>;
+	type ServiceAction = <
+		T = Promise<any>,
+		P extends GenericObject = GenericObject,
+		S = ServiceSettingSchema
+	>(
+		this: Service<S>,
+		params?: P,
+		opts?: CallingOptions
+	) => T;
 
 	interface ServiceActions {
 		[name: string]: ServiceAction;


moonrailgun:
I have no idea how to check `this` type in tsd.

shawnmcknight (approved):
LGTM!

icebob:
It's not breaking old codes?

moonrailgun:
> It's not breaking old codes?

@icebob 

Nope, because old codes definition is not work.

About `this` in typescript. you can check out this reference: https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function

shawnmcknight:
> It's not breaking old codes?

Also, I believe the intent of the original definitions was to do what @moonrailgun has fixed it to do.

icebob:
Ok, thank you!

cs4inctec:
Hello :wave: 
this PR kinda broke our setup, and we are now checking how to update.

--

For context, we provided our own definition of `this` to the `ActionHandler` in order to 
1. get the Moleculer definitions, and
2. get our own values, which were set in the `started` method.

Situation 1 would now work out-of-the-box :+1:, but how to type custom values now?

--

Our workaround now is to define custom values as optional, and force typescript in an `ActionHandler` to ignore the optional definition with `!`.

Workaround Example:
```
type This = Service & {
  readonly client?: AsyncFuntions;
}

async handler(
  this: This,
  ctx: Moleculer.Context<ParamType>,
): Promise<GetType> {
  const result = this.client!.get(56);
  return result;
}
```

ccampanale:
Same thing on my side in that our custom `this` interface used on handlers within the service schema actions, methods, and lifecycle hooks is no longer valid since `this` is set by the framework's TS declaration now and is not assignable to the custom `this` type when set this way.

I played around with the declaration a bit and the best I could come up with is a way of providing a type to `ServiceSchema` and `Service` that defaults to `{}` such that it is not required, not `any`, but also extensible. 

```ts
	type ServiceMethods<S = ServiceSettingSchema, ES = {}> = {
		[key: string]: (
			this: Service<S> & ES,
			...args: any[]
		) => any;
	};
	
	interface ServiceSchema<S = ServiceSettingSchema, ES = {}> {
		name: string;
		version?: string | number;
		settings?: S;
		metadata?: GenericObject;
		mixins?: Array<Partial<ServiceSchema>>;
		methods?: ServiceMethods<S, ES>;
		hooks?: ServiceHooks;

		events?: ServiceEvents;
		created?: ((this: Service<S> & ES) => void) | Array<(this: Service<S> & ES) => void>;
		started?: ((this: Service<S> & ES) => Promise<void>) | Array<(this: Service<S> & ES) => Promise<void>>;
		stopped?: ((this: Service<S> & ES) => Promise<void>) | Array<(this: Service<S> & ES) => Promise<void>>;

		[name: string]: any;
	}
	
	class Service<S = ServiceSettingSchema, ES = {}> implements ServiceSchema {
		constructor(broker: ServiceBroker, schema?: ServiceSchema<S>);

		protected parseServiceSchema(schema: ServiceSchema<S>): void;

		name: string;
		fullName: string;
		version?: string | number;
		settings: S;
		metadata: GenericObject;
		dependencies: string | ServiceDependency | Array<string | ServiceDependency>;
		schema: ServiceSchema<S>;
		originalSchema: ServiceSchema<S>;
		broker: ServiceBroker;
		logger: LoggerInstance;
		actions: ServiceActions;
		Promise: PromiseConstructorLike;
		//currentContext: Context | null;

		_init(): void;
		_start(): Promise<void>;
		_stop(): Promise<void>;

		/**
		 * Wait for the specified services to become available/registered with this broker.
		 *
		 * @param serviceNames The service, or services, we are waiting for.
		 * @param timeout The total time this call may take. If this time has passed and the service(s)
		 * 						    are not available an error will be thrown. (In milliseconds)
		 * @param interval The time we will wait before once again checking if the service(s) are available (In milliseconds)
		 */
		waitForServices(serviceNames: string | Array<string> | Array<ServiceDependency>, timeout?: number, interval?: number, logger?: LoggerInstance): Promise<WaitForServicesResult>;

                [methodName: string]: (
			this: this & ES,
			...args: any[]
		) => any;

		static applyMixins(schema: ServiceSchema): ServiceSchema;
		static mergeSchemas(mixinSchema: ServiceSchema, svcSchema: ServiceSchema): ServiceSchema;
		static mergeSchemaSettings(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaMetadata(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaMixins(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaDependencies(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaHooks(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaActions(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaMethods(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaEvents(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaLifecycleHandlers(src: GenericObject, target: GenericObject): GenericObject;
		static mergeSchemaUnknown(src: GenericObject, target: GenericObject): GenericObject;
	}
``` 

This only goes so far as to address service methods and lifecycle hooks but not actions. 

It also removes `[name: string]: any` from `Service` which may equally as frustrating for some people...

I don't see a clean way of dealing with this but the issue still is that this change has broken the ability to set the `this` type for handlers which means either doing what @cs4inctec suggested or prefix each handler with `// @ts-ignore` which is less than ideal.

moonrailgun:
> Hello 👋
> this PR kinda broke our setup, and we are now checking how to update.
> 
> --
> 
> For context, we provided our own definition of `this` to the `ActionHandler` in order to
> 
> 1. get the Moleculer definitions, and
> 2. get our own values, which were set in the `started` method.
> 
> Situation 1 would now work out-of-the-box 👍, but how to type custom values now?
> 
> --
> 
> Our workaround now is to define custom values as optional, and force typescript in an `ActionHandler` to ignore the optional definition with `!`.
> 
> Workaround Example:
> 
> ```
> type This = Service & {
>   readonly client?: AsyncFuntions;
> }
> 
> async handler(
>   this: This,
>   ctx: Moleculer.Context<ParamType>,
> ): Promise<GetType> {
>   const result = this.client!.get(56);
>   return result;
> }
> ```

Hi, @cs4inctec . I am glad to help you resolve your problem, but i cannot catch your point.
And i cannot reproduce why you have to mark `client` as optional property.

image



This looks like everything is ok.


> I don't see a clean way of dealing with this but the issue still is that this change has broken the ability to set the `this` type for handlers which means either doing what @cs4inctec suggested or prefix each handler with `// @ts-ignore` which is less than ideal.

@ccampanale , i am sad to broken your project, `moleculer` designed in js and not `typescript` friendly, its too **dynamic**. I am try to use origin way and then i give up. I think a clean way to use `moleculer` in `typescript` is not existed, we should wrap it by ourself and restructuring it.

In my mind, if we wanna support extend origin type, definition like `koa` will be fine.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/435abd2a5715373e1518521cb023e17f4689c7a6/types/koa/index.d.ts#L551-L556

Issue #953, opened 2021-06-30
0.14.14 in combination with nats 2.0.8 incorrectly handles "transporter": "nats://server:port" shortcut option


JMS-1:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I am running the latest version [this is the problem]
- [x] I checked the documentation and found no answer [release notes]
- [x] I checked to make sure that this issue has not already been filed [did my best]
- [x] I'm reporting the issue to the correct repository [are quite sure]

## Current Behavior

Before migrating from 0.14.13 and nats 1.X we use the configuration setting "transporter": "nats://server:port". This is no longer working: nats 2.X tries to connect to localhost:4222 (fallback).

## Expected Behavior

nats 2.X should respect the indicated server and port

## Failure Information

Wie analysed the problem: in this case molecular passed the transporter as is as a property **url** to nats. This propery ist respected in nats 1.X but not in nats 2.X where only the **servers** property is inspected. In case of missing servers the fallback ist used.

https://github.com/moleculerjs/moleculer/blob/v0.14.14/src/transporters/nats.js#L33

### Steps to Reproduce

Just try to connect to a remote NATS using the "transporter": "nats://server:port" shortcut configuration and make sure, that no NATS is running on localhost:4222.

### Context

* Moleculer version: 0.14.14
* NodeJS version: 14
* Operating System: Ubuntu 20.10

### Work-Around

Easy but should be noted in the release note as a breaking change - since this configuration works for NATS 1.X as will in prinicple we're fine! 

Or simply change the handling of the "nats://" URL not to just set the url property but the full configuration wirh servers. In my opionion this is the reason this should be considered a bug.

```
"transporter": { "type": "NATS", "options": { "servers": ["server:port"] } } }
```

gautaz:
Hello,

Same issue here :-).

I agree with @JMS-1 considering the current documentation on this matter:
https://moleculer.services/docs/0.14/networking.html#NATS-Transporter

Which displays examples like:
```
module.exports = {
    nodeID: "server-1",
    transporter: "nats://nats.server:4222"
};
```

PR #954, opened 2021-06-30
Issue 953

intech:
add support multiple servers in transport nats: `nats://nats.server1:4222,nats://nats.server2:4222,nats://nats.server3:4222`

#953

### DIFF
--- a/test/unit/transporters/nats.spec.js
+++ b/test/unit/transporters/nats.spec.js
@@ -395,7 +395,7 @@ describe("Tests Nats V2.x", () => {
 		it("check constructor with string param of multiple servers", () => {
 			let transporter = new NatsTransporter("nats://server1:4222,nats://server2:4222,nats://server3:4222");
 			transporter.isLibLegacy = jest.fn(() => false);
-			expect(transporter.opts).toEqual({ preserveBuffers: true, maxReconnectAttempts: -1, servers: ["server1:4222", "nats://server2:4222", "nats://server3:4222"] });
+			expect(transporter.opts).toEqual({ preserveBuffers: true, maxReconnectAttempts: -1, servers: ["server1:4222", "server2:4222", "server3:4222"] });
 		});
 
 		it("check constructor with options", () => {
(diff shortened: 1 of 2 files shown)


icebob (changes requested):
Please add relevant tests

intech:
@icebob this fix works with current unit tests.
I made a quick fix to fix the issue for make a release.
From the tests, you can only add the indication of multiple servers, but now this is critical, it is rather an additional feature that still needs to be documented.

icebob:
I don't think because it should be failed this code: https://github.com/moleculerjs/moleculer/blob/master/test/unit/transporters/nats.spec.js#L392
And currently there is no use-case when the `url` contains multiple servers.

intech:
Yes, after your comment I decided to look through the tests and double-check that everything will be fine and saw this check on the url.
I corrected it and added one more test, but I do it via github ide in the browser and I may not get it the first time, sorry :)

icebob:
Nop, I'm waiting the CI and then I'll merge. Thanks, the quick job.

intech:
Ooook! Got to my computer and did a local test check. :)

test/unit/transporters/nats.spec.js
```diff
@@ -389,13 +389,13 @@ describe("Tests Nats V2.x", () => {
...
 			transporter.isLibLegacy = jest.fn(() => false);
-			expect(transporter.opts).toEqual({ preserveBuffers: true, maxReconnectAttempts: -1, servers: ["nats://localhost"] });
+			expect(transporter.opts).toEqual({ preserveBuffers: true, maxReconnectAttempts: -1, url: "nats://localhost" });
 		});
 
 		it("check constructor with string param of multiple servers", () => {
 			let transporter = new NatsTransporter("nats://server1:4222,nats://server2:4222,nats://server3:4222");
 			transporter.isLibLegacy = jest.fn(() => false);
-			expect(transporter.opts).toEqual({ preserveBuffers: true, maxReconnectAttempts: -1, servers: ["server1:4222", "server2:4222", "server3:4222"] });
+			expect(transporter.opts).toEqual({ preserveBuffers: true, maxReconnectAttempts: -1, url: "nats://server1:4222,nats://server2:4222,nats://server3:4222" });
```
icebob, line 398: But the original issue was that it should be in `servers` and not `url`, wasn't it?

intech: this cannot be done at the constructor level, since the nats version is checked later and therefore I had to transfer the formation of servers to the connection only on nats 2.x
https://github.com/moleculerjs/moleculer/blob/issue-953/src/transporters/nats.js#L147

icebob: ok, you are right.

PR #964, opened 2021-07-15
fix: ActionParamTypes should accept string type

0x0a0d:
### :dart: Relevant issues

```js
export default {
  name: 'test',
  actions: {
    demo: {
      params: {
        a: 'string|max:64'
      }
    }
  }
}
```
Above service config will trigger typescript error
>Type '"string|max:64"' is not comparable to type 'ActionParamTypes'.
because `type ActionParamTypes` missed `string` type (not `"string"`)
### :gem: Type of change

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -65,6 +65,7 @@ declare namespace Moleculer {
 		| "url"
 		| "uuid"
 		| boolean
+		| string
 		| ActionParamSchema;
 	type ActionParams = { [key: string]: ActionParamTypes };
 
@@ -746,6 +747,7 @@ declare namespace Moleculer {
 		 * @param timeout The total time this call may take. If this time has passed and the service(s)
 		 * 						    are not available an error will be thrown. (In milliseconds)
 		 * @param interval The time we will wait before once again checking if the service(s) are available (In milliseconds)
+		 * @param logger the broker logger instance
 		 */
 		waitForServices(serviceNames: string | Array<string> | Array<ServiceDependency>, timeout?: number, interval?: number, logger?: LoggerInstance): Promise<WaitForServicesResult>;
 


0x0a0d:
line 750 just add to fix missing `@param logger` define in jsdoc's warning message

ccampanale:
I'm not entirely sure this is a bug. I believe the current type `ActionParamTypes` assumes use of FastestValidator, the default action param validator that the framework uses. On one hand, adding `string` to the available types may actually make the type checking for correct param types for the validator much less useful as any string will then be valid (making the string literals virtually useless). On the other hand, if you are using a custom validator and not Fastest Validator then the interface for these params may be entirely different.

I'm not sure what the best approach is here but something makes me think that simply adding this type is not it.

Is the example param type valid due to a recent change or is this utilizing a custom validator? The current documentation does support shorthand param types but the example seems to imply more functionality than what is documented.

0x0a0d:
Because the moleculer is using FastestValidator as default validator, so I think default declaration should go with it.
For example, if you use other validator, the default: 'string', 'number', ... may not be accepted by it, you need to write and use your custom declaration.
Adding a type string doesn't make the editor drop the hint
Screen Shot 2021-07-15 at 19 26 32


> Is the example param type valid due to a recent change or is this utilizing a custom validator? The current documentation does support shorthand param types but the example seems to imply more functionality than what is documented.

In the document, I also see an example 
https://moleculer.services/docs/0.14/validating.html#Events-Validation

image

0x0a0d:
> I'm not entirely sure this is a bug

I don't think it's a bug either, but only that option makes sense =))

ccampanale:
@0x0a0d Thanks for the clarification!

I see what you are saying regarding the default declaration and the hints in the editor. I suppose I was thinking less from an intellisense (editor hints) perspective and more from a compile time type checking perspective where this would limit TS's ability to properly check these values, but that's certainly not the end of the world.

I see in the screenshot you provided that there is a comment with a link to the [Shorthand Definitions](https://github.com/icebob/fastest-validator#shorthand-definitions) section in the FastestValidator documentation. This is slick; I honestly didn't know this existed until now. (Probably wouldn't be a bad idea to update the Moleculer docs to point to this, especially in the section that documents the param shorthand capabilities.)

Given that this even works for the default validator I agree that this makes sense as an improvement to the declarations for the framework. Thanks again!

0x0a0d:
@ccampanale you can take a look https://github.com/icebob/fastest-validator
With custom validator (of FastestValidator) 
```js
// custom.validator.js
export const validators = {
  type: 'Fastest',
  options: {
    customRules: {} , // customRule: very power
    defaults: {
      objectID: {
        ObjectID: Types.ObjectId
      }
    },
    aliases: {
      username: 'string|min:5|max:32'
    }
  }
}

// moleculer.config.js
export default {
// ...
validator: validators,
}

// some.service.js
export default {
  actions: {
    login: {
      params: {
        username: 'username' // now we ensure that  in ctx.params.username, it's a string has length from 5 to 32
    }
  }
}
```

ccampanale:
@0x0a0d Thanks for the custom validator example; this also helps to clear things up and makes adding the `string` type to `ActionParamTypes` much clearer.

icebob (approved):
Thanks

PR #966, opened 2021-07-19
d.ts: missing settled on ctx.mcall options

0x0a0d:
change mcall options on context from `CallingOptions` to `MCallCallingOptions`


icebob (approved):
Thx

PR #970, opened 2021-07-22
mcall typing seems to not follow the documentation

### LINKED ISSUE #969: mcall typing seems to not follow the documentation
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository

## Current Behavior
ctx.mcall say it will return a T or Array T, but in fact it depend of the way we pass arguments

image

## Expected Behavior

if I pass an Array  typescript say I will get an array of results
image
and if I pass an object : 
image

## Failure Information

not applicable

### Steps to Reproduce

not applicable

### Reproduce code snippet

not applicable

### Context

* Moleculer version: 0.14.16
* NodeJS version: 14.15.5
* Operating System: windows 21H1 19043.1110

### Failure Logs
not applicable

I will open a PR to fix this, it's just to mark then before .

thib3113:
### :dart: Relevant issues

Fix #969 

### :gem: Type of change

( can be breaking if people miss use T )

with the IDE typing : 
with array :
 image
with an object : 
image

cc @dprslt and @0x0a0d because you have recently updated this code .

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -562,7 +562,8 @@ declare namespace Moleculer {
 		call<T>(actionName: string): Promise<T>;
 		call<T, P>(actionName: string, params: P, opts?: CallingOptions): Promise<T>;
 
-		mcall<T>(def: Array<MCallDefinition> | { [name: string]: MCallDefinition }, opts?: MCallCallingOptions): Promise<Array<T> | T>;
+		mcall<T>(def: Record<string, MCallDefinition>, opts?: MCallCallingOptions): Promise<Record<string, T>>;
+		mcall<T>(def: Array<MCallDefinition>, opts?: MCallCallingOptions): Promise<Array<T>>;
 
 		emit<D>(eventName: string, data: D, opts: GenericObject): Promise<void>;
 		emit<D>(eventName: string, data: D, groups: Array<string>): Promise<void>;
@@ -1089,7 +1090,8 @@ declare namespace Moleculer {
 		call<T>(actionName: string): Promise<T>;
 		call<T, P>(actionName: string, params: P, opts?: CallingOptions): Promise<T>;
 
-		mcall<T>(def: Array<MCallDefinition> | { [name: string]: MCallDefinition }, opts?: MCallCallingOptions): Promise<Array<T> | T>;
+		mcall<T>(def: Record<string, MCallDefinition>, opts?: MCallCallingOptions): Promise<Record<string, T>>;
+		mcall<T>(def: Array<MCallDefinition>, opts?: MCallCallingOptions): Promise<Array<T>>;
 
 		emit<D>(eventName: string, data: D, opts: GenericObject): Promise<void>;
 		emit<D>(eventName: string, data: D, groups: Array<string>): Promise<void>;


index.d.ts
```diff
@@ -562,7 +562,8 @@ declare namespace Moleculer {
 		call<T>(actionName: string): Promise<T>;
 		call<T, P>(actionName: string, params: P, opts?: CallingOptions): Promise<T>;
 
-		mcall<T>(def: Array<MCallDefinition> | { [name: string]: MCallDefinition }, opts?: MCallCallingOptions): Promise<Array<T> | T>;
+		mcall<T>(def: Record<string, MCallDefinition>, opts?: MCallCallingOptions): Promise<Record<string, T>>;
```
shawnmcknight, line 565: The `mcall` syntax is really hard to model with TS.  I think this is on the right path (and a likely improvement over what already is there), but there's a potential for issues.

The generic `T` is, more likely than not, going to be incorrect since each called action in the `mcall` is likely to return a different type.  In its normal form, `mcall` is essentially a `Promise.all` and with the `settled` option its essentially a `Promise.allSettled`.

In looking at the built-in TS definitions for `Promise.all`, they aren't really pretty: https://github.com/microsoft/TypeScript/blob/main/lib/lib.es2015.promise.d.ts#L41

They look like they have streamlined things a bit for the `allSettled` definition (probably because it came along later): https://github.com/microsoft/TypeScript/blob/main/lib/lib.es2020.promise.d.ts#L40

Of course, the big challenge here is that the return type of an action can't be inferred the way most promise results can.  Additionally, we must model an overload with `settled` set to `true`.  I'm not really sure how best to approach this in a satisfactory way.  I'll need to think on it a bit.

Overall, this is definitely an improvement over the existing typings in that the object format has _some_ support.  So I think it would be safe to move forward with this, but maybe we should leave a jsdoc on the type with the current limitations.

Is there an advantage to using `mcall` over an array of `Promise.all([ctx.call, ctx.call, ...])`?  We've always used the latter in our environment.

thib3113: Hello,

I totally agree, but in fact, it's the hard think of a function that will return a generic number of generics objects ...


I never use allSettled, but when checking the type of Promise.all they do a manual typing for 10 results ... 

About the rest, the problem with the actual typing, is if I've an easy use case, like each functions return string, I can't directly use it as an array, because it think it can be an object .

about the more complex ways, in general, I retype the object after getting the objects ( to use my personnal types easier ) .

---

I'm too "new in moleculer" to know if there is a difference between .all of call and mcall ... The doc say to use mcall, so I try mcall 😂

icebob:
What's the status of this PR?

thib3113:
@icebob I don't know ... It seems CI just start (all jobs are queued, waiting) ? so waiting the CI ?

icebob:
@shawnmcknight is it approved by you?

shawnmcknight (approved):
This is definitely an incremental improvement.  Getting complete and accurate typings here would require some more work, but that doesn't diminish the value of adding this update.  LGTM!

PR #980, opened 2021-08-05
Update to use ThisType instead of this override

### LINKED ISSUE #949: Typescript: ThisType in ActionHandler is not work

shawnmcknight:
## :memo: Description

#950 attempted to improve the type inference of action handlers so they knew that the `this` context was `Service`.  However, this led to other scenarios (e.g. middleware definitions) where type errors were occurring because TS was unable to infer the proper `this` value.

This PR reverts the changes from #950 and utilizes a different approach where the `ThisType` utility type is used on the service definition objects.  This allows the properties of those objects to infer the correct value for `this` without adding a type constraint on the methods themselves that leads to other problems.

Note: The use of `ThisType` requires the enabling of `noImplicitThis` or `strict` in the TS compiler options.  It won't cause an error or anything if that is not enabled, but `this` will be inferred as `any` rather than the intended type.

### :dart: Relevant issues

Resolves #949 

### :gem: Type of change

* I've run typechecking before and after this change in a very large repo.  Errors existed prior to the change and no longer exist.
* I tested the original reported issue in #950 to ensure that type inference was still occurring in action handlers for `this`.
* Existing `tsd` tests covered this scenario well and continue to pass.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,10 +44,7 @@ declare namespace Moleculer {
 		trace(...args: any[]): void;
 	}
 
-	type ActionHandler<T = any, S = ServiceSettingSchema> = (
-		this: Service<S>,
-		ctx: Context<any, any>
-	) => Promise<T> | T;
+	type ActionHandler<T = any> = (ctx: Context<any, any>) => Promise<T> | T;
 	type ActionParamSchema = { [key: string]: any };
 	type ActionParamTypes =
 		| "any"
@@ -460,7 +457,7 @@ declare namespace Moleculer {
 		basePath?: string,
 	}
 
-	interface ActionSchema {
+	type ActionSchema<S = ServiceSettingSchema> = {
 		name?: string;
 		rest?: RestSchema | string | string[],
 		visibility?: ActionVisibility;
@@ -476,9 +473,9 @@ declare namespace Moleculer {
 		hooks?: ActionHooks;
 
 		[key: string]: any;
-	}
+	} & ThisType<Service<S>>
 
-	interface EventSchema {
+	type EventSchema<S = ServiceSettingSchema> = {
 		name?: string;
 		group?: string;
 		params?: ActionParams;
@@ -489,9 +486,11 @@ declare namespace Moleculer {
 		context?: boolean;
 
 		[key: string]: any;
-	}
+	} & ThisType<Service<S>>
 
-	type ServiceActionsSchema = { [key: string]: ActionSchema | ActionHandler | boolean; };
+	type ServiceActionsSchema<S = ServiceSettingSchema> = {
+		[key: string]: ActionSchema | ActionHandler | boolean;
+	} & ThisType<Service<S>>;
 
 	class BrokerNode {
 		id: string;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -44,10 +44,7 @@ declare namespace Moleculer {
 		trace(...args: any[]): void;
 	}
 
-	type ActionHandler<T = any, S = ServiceSettingSchema> = (
-		this: Service<S>,
-		ctx: Context<any, any>
-	) => Promise<T> | T;
+	type ActionHandler<T = any> = (ctx: Context<any, any>) => Promise<T> | T;
```
devalexandre, line 47: why remove this:service ?

devalexandre:
When remove `this: Service`, broken some things

index.d.ts
ccampanale: I believe it's because `this` is now set by the `ThisType<Service<s>>` on the entire Action schema on line 476 and similarly for other types that make up the schema.

shawnmcknight: > I believe it's because `this` is now set by the `ThisType<Service<s>>` on the entire Action schema on line 476 and similarly for other types that make up the schema.

That is correct.  The `ThisType` utilization is intended to solve the original problem that adding the `this:Service` was intended to solve, without introducing side effects when you attempt to use an `ActionHandler` (or other type) in a context where `this` isn't `Service`.

It is important to note (as indicated in the original PR comments) that the `noImplicitThis` compiler option must be enabled for `ThisType` to function properly.

shawnmcknight:
> When remove `this: Service`, broken some things

Can you clarify what has been broken?  Please note that the `noImplicitThis` compiler option has to be enabled for `ThisType` to work properly.

I'm happy to look into any new problems that have arisen, but the use of `this:Service` in the function definitions introduced its own set of problems as well, so any solution has to take into account removing those problems while trying to make the original problem that adding `this:Service` was intended to fix go away.  This solution was the only solve I found that satisfied both use cases, but it does require that additional compiler option.

icebob:
@devalexandre any news?

shawnmcknight:
@icebob I think this is fine.  @devalexandre had responded in https://github.com/moleculerjs/moleculer-template-project-typescript/issues/37#issuecomment-894313335 that the change was working (provided that `noImplicitThis` was turned on).  I suppose a PR needs to be made against that repo to turn on that setting though.

PR #985, opened 2021-08-19
FIX: only set heartbeatInterval + heartbeatTimeout in registry will not work

roytan883:
## :memo: Description

FIX: only set heartbeatInterval + heartbeatTimeout in registry will not work

### :scroll: Example code
```js
if (now - node.lastHeartbeatTime > this.broker.options.heartbeatTimeout) {
//this will always be true when this.broker.options.heartbeatTimeout not set
``` 

### FIX:
should use 
```js
if (now - node.lastHeartbeatTime > this.opts.heartbeatTimeout) {
```


roytan883:
I got a very strange bug: no matter what `heartbeatTimeout` i set in `registry.discoverer.options`. It always warning `Heartbeat is not received from 'xxxxxx' node.` 
Then i found the reason is : i removed `heartbeatTimeout` in broker options. 
`heartbeatTimeout` set in one place should work. This pull should fix it.

PR #987, opened 2021-08-19
Fixes 978

### LINKED ISSUE #978: Cacher should be ignored by the action handler if cacher is not connected
**Is your feature request related to a problem? Please describe.**
Currently if the cacher for a service is not in a healthy state, the service will fail to respond to action calls as it's trying to handle the action from the underlying cache associated with the broker.

This isn't ideal in a microservice world as it creates a strict dependency between the service and the cache.

If this weren't the case, it would allow the developer to write much more fault tolerant services.

**Describe the solution you'd like**
I propose a `connected` flag should be implemented by the cachers. Then the handler middleware could check the connected state of the cacher before attempting to get/set, if `connected === false` the request should be directed straight to the handler and the response should not be set in the cache.

There will be some cases where the service is strictly dependent on the underlying broker cacher. However, i believe it should be the responsibility of the developer to decide whether their service is healthy when the cacher is offline on a service by service basis.

AndreMaz:
### :dart: Relevant issues

Fixes #978 

### :gem: Type of change

### DIFF
--- a/src/cachers/base.js
+++ b/src/cachers/base.js
@@ -32,7 +32,7 @@ class Cacher {
 		});
 
 		/** @type {boolean} Flag indicating the connection status */
-		this.connected = false;
+		this.connected = null; // Init as null for backward compatibility
 	}
 
 	/**
@@ -336,7 +336,7 @@ class Cacher {
 
 					// Cache is enabled but not in healthy state
 					// More info: https://github.com/moleculerjs/moleculer/issues/978
-					if (!this.connected) {
+					if (this.connected === false) {
 						this.logger.debug(
 							"Cacher is enabled but it is not connected at the moment... Calling the handler"
 						);
--- a/src/cachers/base.js
+++ b/src/cachers/base.js
@@ -28,9 +28,11 @@ class Cacher {
 		this.opts = _.defaultsDeep(opts, {
 			ttl: null,
 			keygen: null,
-			maxParamsLength: null,
-			connected: false
+			maxParamsLength: null
 		});
+
+		/** @type {boolean} Flag indicating the connection status */
+		this.connected = false;
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


icebob (approved):

icebob:
@JordanPawlett Is it what you think?

src/cachers/base.js
```diff
@@ -30,6 +30,9 @@ class Cacher {
 			keygen: null,
 			maxParamsLength: null
 		});
+
+		/** @type {boolean} Flag indicating the connection status */
+		this.connected = false;
```
JordanPawlett, line 35: Maybe `this.connected` should be `undefined` by default on the base cacher.

This would avoid breaking any Custom-Cachers currently in use, if the check below check if `this.connected` was `undefined && false`

JordanPawlett:
> @JordanPawlett Is it what you think?

Yes, exactly what I had in mind. Should allow service's to respond to RPC calls successfully even when the cacher has failed.
My only worry would be breaking existing custom-cacher implementation that don't have `this.connected` implemented.

src/cachers/base.js
icebob: @AndreMaz please initialize this to `null` and set `this.connected = false` in Redis `init` method.

src/cachers/base.js
```diff
@@ -331,6 +334,15 @@ class Cacher {
 					// Disable caching with `ctx.meta.$cache = false`
 					if (ctx.meta["$cache"] === false) return handler(ctx);
 
+					// Cache is enabled but not in healthy state
+					// More info: https://github.com/moleculerjs/moleculer/issues/978
+					if (!this.connected) {
```
icebob, line 339: Please change it to `if (this.connected === false)`

icebob (changes requested):

icebob (approved):
Thanks @AndreMaz, you can merge it.

Issue #861, opened 2021-01-26
Keep Moleculer error classes for calls between services


gautaz:
**Is your feature request related to a problem? Please describe.**
From my understanding, the current Moleculer behavior when throwing an exception from a service B called by a service A is for the service A to receive the error as a generic Javascript [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) instance.
Whether the original error is a [MoleculerError, MoleculerClientError, MoleculerServerError](https://moleculer.services/docs/0.14/errors.html) or a custom error inheriting from the previous ones does not seem to change this behavior.

**Describe the solution you'd like**
Service A cannot know of custom error classes defined in service B but at least it is aware of all Moleculer error classes.
From service A point of view, it would help to distinguish client errors from server errors, this could be handled by receiving a MoleculerClientError or a MoleculerServerError depending on what the original error from service B inherited from.
In fact, that was what one of my developers expected naturally.

**Describe alternatives you've considered**
AFAIK we can only rely on error names, type codes or any other mean detailed in the error content.
I did not find for now a native "generic way" to know if the error is due to the caller or the callee.

**Additional context**
N/A

icebob:
The `Transit` tries to recreate the original error based on the error name here:
https://github.com/moleculerjs/moleculer/blob/498b008c1e8bcf6049f6bd762c0015d4c47f6100/src/errors.js#L389-L415

Could you create a minimal reproduce example the we can investigate it?

gautaz:
@icebob Thanks for your answer.

I guess the issue here is more about documentation as the behavior you describe should be detailed in https://moleculer.services/docs/0.14/errors.html.

In particular in the section [Create custom errors](https://moleculer.services/docs/0.14/errors.html#Create-custom-errors), the documentation should perhaps emphasize the fact that the name is used later on to unmarshall errors.

What is also perhaps bothering for developers is that they cannot set their own error names without crippling the error unmarshalling.
Perhaps a field name that clarifies the intention (something like "moleculerErrorClass" instead of "name") would help understanding its effect without looking at the documentation.

gautaz:
@icebob The problem can easily be reproduced:

```js
const { MoleculerClientError } = require("moleculer").Errors;

class CustomClientError extends MoleculerClientError {
    constructor(msg, data) {
        super(msg || `This is a custom client error.`, 400, "CUSTOM_CLIENT_ERROR", data);
    }
}

[...]

PR #1017, opened 2021-10-04
 Keep custom error classes for calls between services

FFKL:
## :memo: Description
I'm working on implementing a solution to preserve custom error classes between services (see Relevant issues). I've decided to add `recreateError` option to `ServiceBroker` options and call this function after the built-in one.

### :dart: Relevant issues
https://github.com/moleculerjs/moleculer/issues/861

### :gem:  Type of change

### :scroll: Example code
```js
const customRecreateError = function (err) {
			switch (err.name) {
				case "MyCustomError":
					return new MyCustomError(err.name, err.code, err.type, err.data);
			}
		};
			logger: false,
			nodeID: "node1",
			recreateError: customRecreateError,
			transporter: new FakeTransporter()
		});
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -844,6 +844,7 @@ declare namespace Moleculer {
 		cacher?: boolean | Cacher | string | GenericObject;
 		serializer?: Serializer | string | GenericObject;
 		validator?: boolean | BaseValidator | ValidatorNames | ValidatorOptions;
+		errorRegenerator?: Errors.Regenerator;
 
 		metrics?: boolean | MetricRegistryOptions;
 		tracing?: boolean | TracerOptions;
@@ -1034,6 +1035,7 @@ declare namespace Moleculer {
 		cacher?: Cacher;
 		serializer?: Serializer;
 		validator?: BaseValidator;
+		errorRegenerator?: Errors.Regenerator;
 
 		tracer: Tracer;
 
@@ -1434,6 +1436,8 @@ declare namespace Moleculer {
 
 		interface PlainMoleculerError extends MoleculerError {
 			nodeID?: string;
+
+			[key: string]: any;
 		}
 
 		class Regenerator {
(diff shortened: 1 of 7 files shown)


FFKL:
@icebob I want you to look at this draft issue. I also thinking about using of custom error serializer (as `serializer` with base class). It will help a programmer to flexibly customize it's errors, use custom plain error objects for sending and resolve name conflicts with the default ones. Also it could be really helpful to infer class types in error handlers with TypeScript - `instanceof MyError`. What do you think?

icebob:
Thanks, the `recreateError` broker option solution is acceptable for me.

Could you give an example what do you mean error serializer?

FFKL:
@icebob Maybe it's not a `Packer`;) But you see the main idea.
```js
class BaseErrorPacker {
	pack(plainError, sender) {
		const Class = module.exports[plainError.name];
		if (Class) {
			switch (plainError.name) {
				// ... E.recreateError switch;
			}
		}
                return err ? err : this.createDefaultError(err); // new Error(err);
	}
	unpack(error) {
		return {
			data: err.data,
			name: err.name
			// ...etc
		};
	}
}
```
Usage:
```js
_createErrFromPayload(error, sender) {
	return this.errorPacker.pack(error, sender);
}

_createPayloadErrorField(err) {
	return this.errorPacker.unpack(err);
}
```
In client code:
```js
class MyErrorPacker extends BaseErrorPacker {
	pack(plainError, sender) {
		switch (plainError) {
			// build custom error
		}
		return super.pack(plainError, sender);
	}
	unpack(error) {
		return {
			data: err.data,
			name: err.name,
			reqId: err.reqId
			// ...etc
		};
	}
}

	nodeID: "node1",
	errorPacker: new MyErrorPacker()
});
```

icebob (approved):
Good job!

PR #1021, opened 2021-10-13
fix(transporters/nats): parse user and password of server urls

savsgio:
## :memo: Description

Now the nats transporter does not parse the user and password from the url, just take and set the server hosts.
This PR fixes that, taking the first user and password found in the servers.

### :gem: Type of change

### :scroll: Example code
```js
new NatsTransporter("nats://MYUSER:MYPASS@server1:4222");
``` 

- [x] Test NatsTransporter connect & disconnect & reconnect -> check connect options servers

### DIFF
--- a/test/unit/transporters/nats.spec.js
+++ b/test/unit/transporters/nats.spec.js
@@ -507,7 +507,7 @@ describe("Tests Nats V2.x", () => {
 
 		beforeEach(() => {
 			transporter = new NatsTransporter(
-				"nats://server1:4222,nats://server2:4222,nats://server3:4222"
+				"nats://myuser:mypass@server1:4222,nats://server2:4222,nats://server3:4222"
 			);
 			transporter.isLibLegacy = jest.fn(() => false);
 			transporter.init(transit, msgHandler);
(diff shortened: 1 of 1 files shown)


icebob:
It means different auth for servers is not available in nats.js? E.g. `"nats://user1:pass1@server1:4222,nats://user2:pass2@server2:4222,nats://user3:pass3@server3:4222"`

oliversalzburg:
Just FYI, I believe this introduced a regression.

Before, it was possible to use `url`+`user`+`pass` to connect to your NATS server. This change removes the authentication information and ultimately kills the entire application.

intech:
In this implementation, we also miss the ability to set a token. I suggest doing this as an additional check, if only user is passed and there is no password, then we use the token.

AndreMaz:
Ups! We're terribly sorry for the inconvenience @oliversalzburg Could you please create a PR with the fix?

oliversalzburg:
Don't worry :) I kinda feel like it was a configuration mistake on our side anyway. The idea is probably to use servers+user+pass _or_ URL with all details included.

Maybe a validation in this regard could help to prevent others from making the same mistake?

PR #1004, opened 2021-09-09
add cache keygen to service action

0x0a0d:
allow define `keygen` in `action.cache` that override `cacher.defaultKeygen` and `cacher.keygen`
use case
```js
const serviceSchema = {
  actions: {
    find: {
      cache: {
        keygen(actionName, params, meta, defaultKeys) {
          // add meta.user._id to defaultKeys
          return this.defaultKeygen(actionName, params, meta, ['#user._id'].concat(defaultKeys))
        }
      }
    }
  }
}
```

PR #1025, opened 2021-10-28
d.ts: Cache keygen

0x0a0d:
+ add typescript for #1004
+ fix some typescript `ban-types`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -427,12 +427,12 @@ declare namespace Moleculer {
 	}
 
 	type ActionCacheEnabledFuncType = (ctx: Context<any, any>) => boolean;
-	type ActionCacheKeygenFuncType<P = Record<string, any>, M = any, K = any[]|null> = (actionName: string, params: P, meta: M, defaultKeys: K) => string;
 
 	interface ActionCacheOptions<P = Record<string, any>, M = any, K = any[]|null> {
 		enabled?: boolean | ActionCacheEnabledFuncType;
 		ttl?: number;
-		keys?: string[] | ActionCacheKeygenFuncType<P, M, K>;
+		keys?: Array<string>;
+		keygen?: CacherKeygenFuncType<P, M, K>;
 		lock?: {
 			enabled?: boolean;
 			staleTime?: number;
@@ -1196,9 +1196,10 @@ declare namespace Moleculer {
 		deserialize(type: string, data: Buffer): Packet;
 	}
 
+	type CacherKeygenFuncType<P = Record<string, any>, M = any, K = any[]|null> = (actionName: string, params: P, meta: M, defaultKeys: K) => string;
 	interface CacherOptions {
 		ttl?: number;
-		keygen?: Function;
+		keygen?: CacherKeygenFuncType;
 		maxParamsLength?: number;
 		[key: string]: any;
 	}
--- a/index.d.ts
+++ b/index.d.ts
@@ -673,7 +673,7 @@ declare namespace Moleculer {
 		version?: string | number;
 		settings?: S;
 		dependencies?: string | ServiceDependency | Array<string | ServiceDependency>;
-		metadata?: GenericObject | any;
+		metadata?: any;
 		actions?: ServiceActionsSchema;
 		mixins?: Array<Partial<ServiceSchema>>;
 		methods?: ServiceMethods;
--- a/index.d.ts
+++ b/index.d.ts
@@ -22,7 +22,7 @@ declare namespace Moleculer {
 		init(opts: LoggerConfig | Array<LoggerConfig>): void;
 		stop(): void;
 		getLogger(bindings: GenericObject): LoggerInstance;
-		getBindingsKey(bindings: GenericObject): String;
+		getBindingsKey(bindings: GenericObject): string;
 
 		broker: ServiceBroker;
 	}
@@ -251,7 +251,7 @@ declare namespace Moleculer {
 
 		lastSnapshot: GenericObject | null;
 		dirty: boolean;
-		values: Map<String, GenericObject>;
+		values: Map<string, GenericObject>;
 
 		constructor(opts: BaseMetricOptions, registry: MetricRegistry);
 		setDirty(): void;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -520,7 +521,7 @@ declare namespace Moleculer {
 		disconnected(): void;
 	}
 
-	class Context<P = unknown, M extends object = {}> {
```
icebob, line 523: Please explain why you change it and it can break other TS users' projects.

shawnmcknight: This was made `unknown` a while ago because with a default of `any` you, by default, get no type safety with moleculer context.

index.d.ts
```diff
@@ -427,11 +427,12 @@ declare namespace Moleculer {
...
 
 	type ActionCacheEnabledFuncType = (ctx: Context<any, any>) => boolean;
+	type ActionCacheKeygenFuncType<P = Record<string, any>, M = any, K = any[]|null> = (actionName: string, params: P, meta: M, defaultKeys: K) => string;
 
-	interface ActionCacheOptions {
+	interface ActionCacheOptions<P = Record<string, any>, M = any, K = any[]|null> {
 		enabled?: boolean | ActionCacheEnabledFuncType;
 		ttl?: number;
-		keys?: Array<string>;
+		keys?: string[] | ActionCacheKeygenFuncType<P, M, K>;
```
shawnmcknight, line 435: I must be missing something, but I don't see anywhere that `keys` can accept anything other than a `string[]`.  What code is checking this property for a function?  In the referenced PR I see that it allowed specification of a `keygen` property, but I'm not sure where you can override the value of `keys` with a function.

index.d.ts
```diff
@@ -672,7 +673,7 @@ declare namespace Moleculer {
 		version?: string | number;
 		settings?: S;
 		dependencies?: string | ServiceDependency | Array<string | ServiceDependency>;
-		metadata?: GenericObject;
+		metadata?: GenericObject | any;
```
shawnmcknight, line 676: It seems that metadata _can_ be `any` type (per the docs) and doesn't need to be an object.  With that said, a type union with `any` evaluates to `any`, so the `GenericObject` can be eliminated from this union.

0x0a0d:
@shawnmcknight it was accepted as a function because of my #1004 pull

index.d.ts
```diff
@@ -1523,8 +1524,8 @@ declare namespace Moleculer {
 		show(): this;
 		hide(): this;
 		find(command: string): Vorpal.Command;
-		exec(command: string): Promise<{}>;
-		execSync(command: string): Promise<{}>;
+		exec(command: string): Promise<any>;
```
shawnmcknight, line 1527: I agree that `{}` is not a good type to use here, but I'm also unclear if `any` is appropriate.  I can't find any reference to `Vorpal` in moleculer so I'm not sure why it's included here at all.  I know `moleculer-repl` is based on the Vorpal library, but I can't find any reference to `exec` or `execSync` in either `moleculer-repl` or Vorpal so I'm totally unsure about this.

`any` is safe so I guess I'm okay with it, but I just don't know why its here at all.

index.d.ts
```diff
@@ -440,7 +441,7 @@ declare namespace Moleculer {
 
 	type ActionVisibility = "published" | "public" | "protected" | "private";
 
-	type ActionHookBefore = (ctx: Context<any, any>) => Promise<void> | void;
+	type ActionHookBefore = (ctx: Context<any, any>) => Promise<any> | any;
```
shawnmcknight, line 444: @icebob Can before hooks return anything?  Its not documented that they can (after hooks explicitly say they can return something).  If I'm understanding `action-hooks.js` properly, it doesn't seem like the value resolved from a before hook would do anything at all.  I'd hesitate to make this change unless there is a specific reason for returning a value from a before hook.

shawnmcknight (changes requested):
Thanks for the PR.  Some of these changes look good, but I've got some questions on others than I'm not sure about.  Please let me know your thoughts.

index.d.ts
icebob: Yeah, it has no effect, return value is not processed.

index.d.ts
icebob: Yeah, it came with #1004 pull as a new feature.

icebob: @0x0a0d Shawn has right, the #1004 adds `keygen` property to the action `cache` definition and not change the `keys`.

icebob (changes requested):

icebob (approved):

icebob:
@shawnmcknight could you make a last check, please?

index.d.ts
```diff
@@ -428,10 +428,11 @@ declare namespace Moleculer {
 
 	type ActionCacheEnabledFuncType = (ctx: Context<any, any>) => boolean;
 
-	interface ActionCacheOptions {
+	interface ActionCacheOptions<P = Record<string, any>, M = any, K = any[]|null> {
```
shawnmcknight, line 431: The `keys` property is indicated as `Array<string> | undefined` but the `K` generic here (as well as in `CacherKeygenFuncType` defaults to `any[]`.  Either `keys` needs to be widened (which I don't think is a good idea) or the `K` generic should be eliminated and the `defaultKeys` parameter set to `defaultKeys?: string[]`.

Although the jsdocs indicate that `keys` can be `null`, I don't actually see that.  Based on the types below, keys is either `undefined` or `string[]` so I think the `null` would be incorrect which is resolved with my suggestion above.

index.d.ts
```diff
@@ -1195,9 +1196,10 @@ declare namespace Moleculer {
 		deserialize(type: string, data: Buffer): Packet;
 	}
 
+	type CacherKeygenFuncType<P = Record<string, any>, M = any, K = any[]|null> = (actionName: string, params: P, meta: M, defaultKeys: K) => string;
```
shawnmcknight, line 1199: The `Type` suffix here is probably extraneous.  We get the same benefit from naming this `CacherKeygenFunc` since we already know its a type.

shawnmcknight: My personal preference is to use `unknown` in place of `any` whenever any generic is getting defaulted since `any` becomes a free-for-all without any type safety.  If a type has not been provided explicitly in a generic or inferred by TS, treating it as `any` becomes risky.

shawnmcknight (changes requested):
Thanks for the updates.  I had a few other suggestions to tighten things up a bit more.

0x0a0d:
@shawnmcknight you're right

0x0a0d:
@shawnmcknight you're right

shawnmcknight (approved):

icebob:
Thanks @shawnmcknight!

@0x0a0d in the future, please create small and exact PRs. As you see PRs that contain multiple changes can cause problems at reviewing and take a longer time.

PR #1029, opened 2021-11-12
add PromiseConstructorPolyfill

0x0a0d:
maybe this resolve #1028


0x0a0d:
removed
client may not use native Promise

Issue #1026, opened 2021-11-04
how to approach mutually exclusive dependencies


OutdatedVersion:
I'm having trouble finding a supported way to do mutually exclusive dependencies with Moleculer's built-in service [dependency API](https://moleculer.services/docs/0.14/services.html#Dependencies).

For example, the snippet below has two implicit, unasserted, dependencies (`v2.service` and `v1.service`) -- it would be nice to have more confidence that at least one of those is available at runtime.
```ts
const things = await ctx.call(
 'v2.service.getThings',
 { active: true },
 {
    fallbackResponse: () => {
       return ctx.call('v1.service.getThings', { active: true });
    },
  }
);
```

I, without looking at the code/examples, tried a `dependencies` of `['service']` and `[{ name: 'service' }]` thinking Moleculer may do some sort of resolution to match the name... with no luck. From looking around, I believe support would require a Moleculer change. 

Something like:
```ts
{
   dependencies: [
     { or: ['v1.service', 'v2.service']  }
   ],
}
```
could facilitate some special logic in the block below for "one or the other" searches
https://github.com/moleculerjs/moleculer/blob/b71bd71dc2375a72d32c25e2718b4b36a8b57f60/src/service-broker.js#L1022-L1027

It seems best for members of the `or` special case to support the full set of version keys (shorthand with/without version, object, etc.). Though, I believe that would require enriching the `Registry#hasService` API to do partial matches. Not really sure on that one.

Some questions:
- [ ] Are these assumptions correct or is there already a way to do this?
- [ ] If not, is there a particular path forward you think would be best?

icebob:
You can't do that currently with `dependencies`. But the idea is good, but I suggest other solution than `or: []` because I think there is no real use-case that somebody wants to wait for different services with `or`. 
My suggestion is that version can be an array, like:
```js
{
   dependencies: [
     { name: "myService", version: [1, 2] }
   ],
}
```

icebob:
Current workaround if you don't use the `dependencies` in the schema, instead use the `waitForServices` in `started`. E.g.
```js
{
    name: "otherService",

    started() {
        while (true) {
            try {
                await this.waitForServices("v1.myService", 1000);
                break;
            } catch(err) {
                try {
                    await this.waitForServices("v2.myService", 1000);
                    break;
                } catch(err) {
                    // Wait again
                }
            }
        }
    }
}
```

PR #1030, opened 2021-11-12
add support for either-or versions to waitForServices

OutdatedVersion:
## :memo: Description

Add support for a dependency formatted as `{ name: 'z', versions: [1, 2] }` - either `v1.z` or `v2.z` will satisfy this.

### :dart: Relevant issues

#1026 

### :gem: Type of change

### DIFF
--- a/src/service-broker.js
+++ b/src/service-broker.js
@@ -1005,42 +1005,68 @@ class ServiceBroker {
 		serviceNames = _.uniq(
 			_.compact(
 				serviceNames.map(x => {
-					if (utils.isPlainObject(x) && x.name)
-						return this.ServiceFactory.getVersionedFullName(x.name, x.version);
-
-					if (utils.isString(x)) return x;
+					if (utils.isPlainObject(x) && x.name) {
+						if (Array.isArray(x.version)) {
+							return x.version.map(v =>
+								this.ServiceFactory.getVersionedFullName(x.name, v)
+							);
+						} else {
+							return this.ServiceFactory.getVersionedFullName(x.name, x.version);
+						}
+					} else if (utils.isString(x)) {
+						return x;
+					}
 				})
 			)
 		);
 
 		if (serviceNames.length == 0) return this.Promise.resolve({ services: [], statuses: [] });
 
-		logger.info(`Waiting for service(s) '${serviceNames.join(", ")}'...`);
+		logger.info(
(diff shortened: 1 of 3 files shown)


src/service-broker.js
```diff
@@ -986,36 +986,85 @@ class ServiceBroker {
 	waitForServices(serviceNames, timeout, interval, logger = this.logger) {
 		if (!Array.isArray(serviceNames)) serviceNames = [serviceNames];
 
-		serviceNames = _.uniq(
+		class ServiceToken {
```
icebob, line 989: I feel the PR is a little complicated. Please try to solve it without sub-classes. Regarding performance, it's not optimal to create sub-classes and instances of classes in a method.

src/service-broker.js
```diff
@@ -986,36 +986,85 @@ class ServiceBroker {
...
+			constructor(name, version) {
+				this.name = name;
+				this.version = version;
+			}
+
+			get fullName() {
+				if (this.version) {
+					return `v${this.version}.${this.name}`;
```
icebob, line 997: It's not that simple. Instead use `Service.getVersionedFullName`

test/unit/service-broker.spec.js
```diff
@@ -1709,7 +1709,7 @@ describe("Test broker.waitForServices", () => {
 					services: ["posts"],
 					statuses: [{ name: "posts", available: true }]
 				});
-				expect(broker.registry.hasService).toHaveBeenCalledTimes(6);
+				expect(broker.registry.hasService).toHaveBeenCalledTimes(12);
```
icebob, line 1712: Why is duplicated the number of calls?

icebob (changes requested):

icebob:
Are you going to work on it?

src/service-broker.js
OutdatedVersion: Class based was the cleanest/simplest solution I found working on this - I toyed with a pure object/function but that got out of hand imo. Do you have any suggestions for a different approach? Thanks for your time here

OutdatedVersion: @icebob Would you be comfortable moving forward with this approach if we found a better place to store these classes?

OutdatedVersion:
@icebob This has been reworked, if you have a moment to look. Thanks!

icebob (approved):
I like it, thanks!

PR #1032, opened 2021-11-16
add missing maxSafeObjectSize broker option

FFKL:
## :memo: Description

Add  `maxSafeObjectSize` broker option to `index.d.ts` according to https://github.com/moleculerjs/moleculer/blob/7bd4d72d05e79756654638cb449c471a9302717b/src/service-broker.js#L125-L132


icebob (approved):
Good catch

PR #1035, opened 2021-11-19
fix: Add ThisType for remove error in typescript template

devalexandre:
### :gem: Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### :scroll: Example code
```ts


``` 

When using some Adapter with typescript support we have to use ThisType in the mixin , that way when injecting in the service mixin, it shows an error , because the mixins type in the ServiceSchema interface, allows only mixins being Array<Partial<ServiceSchema>>.

This restriction limits the implementation of more typescript adapters.

image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -675,7 +675,7 @@ declare namespace Moleculer {
 		dependencies?: string | ServiceDependency | Array<string | ServiceDependency>;
 		metadata?: any;
 		actions?: ServiceActionsSchema;
-		mixins?: Array<Partial<ServiceSchema>> | Array<Partial<ServiceSchema>, ThisType<Service>>;
+		mixins?: Array<Partial<ServiceSchema>> | Array<Partial<ServiceSchema> & ThisType<Service>>;
 		methods?: ServiceMethods;
 		hooks?: ServiceHooks;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -675,7 +675,7 @@ declare namespace Moleculer {
 		dependencies?: string | ServiceDependency | Array<string | ServiceDependency>;
 		metadata?: any;
 		actions?: ServiceActionsSchema;
-		mixins?: Array<Partial<ServiceSchema>>;
+		mixins?: Array<Partial<ServiceSchema>> | Array<Partial<ServiceSchema>, ThisType<Service>>;
 		methods?: ServiceMethods;
 		hooks?: ServiceHooks;
 


icebob (changes requested):
CI test is broken.

index.d.ts
```diff
@@ -675,7 +675,7 @@ declare namespace Moleculer {
 		dependencies?: string | ServiceDependency | Array<string | ServiceDependency>;
 		metadata?: any;
 		actions?: ServiceActionsSchema;
-		mixins?: Array<Partial<ServiceSchema>>;
+		mixins?: Array<Partial<ServiceSchema>> | Array<Partial<ServiceSchema>, ThisType<Service>>;
```
shawnmcknight, line 678: `Partial<T>` only accepts one generic value, so this is not a valid type.

I'm not really understanding the use case with `ThisType` in place.  A mixin is really just a partial service schema that gets merged together with the main service schema.  What problem is this trying to solve?

shawnmcknight (changes requested):
The use of `Partial` is invalid, but I also don't really understand the problem that you are attempting to solve here

index.d.ts
devalexandre: I try remove ThisType and run

devalexandre: @shawnmcknight  if not use ThisType, have use it.

image


but not work because, aren't implemented.

devalexandre: image

shawnmcknight: @devalexandre `this.DbMixin` is set to equal the return value of the `start()` method of your `DbConnection` class.  What is `start()` returning?  The error message is implying that `start()` is returning `void` which makes sense for a `start()` method generally speaking, but doesn't work for what a mixin should be.

A mixin is really just an object that represents part of a service's schema that gets merged into the mixed service's schema.  While a class method could certainly return that object, I've usually either seen a mixin as a pure object representing a partial service schema (such as shown in the [docs example](https://moleculer.services/docs/0.14/services.html#Mixins)) or an invoked function that returns an object representing a partial service schema.  Your class seems to represent a service that you want to reuse, but that's not really how mixins are intended to work AFAIK.

shawnmcknight: @devalexandre I also noticed that you're using `Service.mergeSchemas` which is a static method which is intended to merge a mixin into the mixed service.  It gets called automatically to merge the declared mixins when calling `parseServiceSchema`.  Honestly, I'm not even sure that some of these static member should be expected to be used by external consumers.  I would anticipate that those should really be considered private members and not really intended for external use.  I see they have been added to the TS typing, but my instinct is that is a mistake.  @icebob What do you think there?

devalexandre: @shawnmcknight  So the best way would be to use the mixin pattern, and remove the use of static methods and the use of partials?

https://github.com/moleculerjs/moleculer-template-project-typescript/blob/master/template/mixins/db.mixin.ts.

in the example given in the typescript template repository, this way also using ThisType and static method, I also think this is bad practice but as it is the default used I kept it that way, to avoid big changes

devalexandre:
I'm going to close this PR, as the  @shawnmcknight  it doesn't make sense to use ThisType

index.d.ts
icebob: > @icebob What do you think there?

yeah, the `merge...` static methods are only for internal usages.

Issue #720, opened 2020-04-16
List of actions to be excluded from tracing


ErmakovSergey:
Can you add exclude feature for **tracing** similar to **metrics**?

intech:
+1

synologic:
Fairly simple i believe,

in tracer.js, change spanFinished function:

```
spanFinished(span) {
                if (span.sampled) {
                        if (this.opts.exclude && this.opts.exclude.length != 0 && this.opts.exclude.indexOf(span.tags.action.name) != -1) return;
                        this.invokeExporter("spanFinished", [span]);
                }
        }
```
then in the tracing config, add an array 

`exclude : ["action1","action2"]`

This will apply for all tracers, same logic can be applied individually.

Alternatively/additionally one can filter by span.name as well (eg this would apply to ApiGw to filter by REST action as span.tags.action.name in ApiGw is api.rest)

AndreMaz:
How this feature would work?
Imagine the following scenario:
![image](https://user-images.githubusercontent.com/9802754/187738222-eb667820-2e1e-464c-9290-f4c5e4450676.png)
and `excludes` field with [`"math.**"`]

In this case the `excludes` means that if we call `math.random` directly there won't by any tracing? 
Or `excludes` means that if we call `greeter.hello` everything after `math.random`, including `math.random`, is not part of the span? Like so:
![image](https://user-images.githubusercontent.com/9802754/187739116-b69e3670-d005-4025-8a61-d55128011b7d.png)

intech:
@AndreMaz I think it should only be for `math.*`
![tracing](https://user-images.githubusercontent.com/451306/187750231-838466a2-28c5-462b-817c-8e3dca97280b.jpg)

AndreMaz:
So, under the hood, it is still necessary to collect all spans and only when pushing or printing them the `mat.**` is excluded?

intech:
Yes, we have discussed this before in discord. In the current implementation, it is not possible. Even if tracing is disabled, it will still collect spans and won't send them.
It is now an acceptable solution to get the functionality to look for unnecessary services from the trace.

PR #1059, opened 2022-01-23
Introduces exclusions from tracing

synologic:
## :memo: Description

Introduces support for excluding actions from tracing.
In the tracing configuration object, two arrays can be added:

exclude : ["service.action","service1.action1"] - will exclude by service and action name
excludeByRest : ["GET /service/action"] - Useful to exclude REST actions from tracing within API Gateway

This will apply gobally, to any exporters configured
### :dart: Relevant issues
#720 

### :gem: Type of change

- [x ] New feature (non-breaking change which adds functionality)

### :scroll: Example code
```js
    tracing: {
        enabled: true,
        exclude : ["service.action"],
        excludeByRest : ["GET /service/action"],
        // Available built-in exporters: "Console", "Datadog", "Event", "EventLegacy", "Jaeger", "Zipkin"
        exporter: {
            type: "Console", // Console exporter is only for development!
            options: {
                // Custom logger
                logger: null,
                // Using colors
                colors: true,
                // Width of row
                width: 100,
                // Gauge width in the row
                gaugeWidth: 40
            }
        }
    },

``` 

- [x ] I have performed a self-review of my own code

### DIFF
--- a/src/tracing/tracer.js
+++ b/src/tracing/tracer.js
@@ -318,9 +318,8 @@ class Tracer {
 	 */
 	spanFinished(span) {
 		if (span.sampled) {
-			if (this.opts.exclude && this.opts.exclude.length != 0 && this.opts.exclude.indexOf(span.tags.action.name) != -1) return;
-			if (this.opts.excludeByRest && this.opts.excludeByRest.length != 0 && this.opts.excludeByRest.indexOf(span.name) != -1) return;
-			
+			if (this.opts.exclude && this.opts.exclude.length !== 0 && this.opts.exclude.indexOf(span.tags.action.name) !== -1) return;
+			if (this.opts.excludeByRest && this.opts.excludeByRest.length !== 0 && this.opts.excludeByRest.indexOf(span.name) !== -1) return;
 			this.invokeExporter("spanFinished", [span]);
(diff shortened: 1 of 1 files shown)


AndreMaz:
Hi @synologic can you please add the relevant tests?

icebob:
Thanks for your work, but unfortunately I should close it, because it's too far from what I planned for this feature.

1. I don't support this excludeByRest option. This is the core moleculer module, which has no REST feature. REST is implemented in moleculer-web. So logically it's not a good place.
2. This `exclude` logic is too simple. If you check the other modules which have include/exclude features, all of them uses the `match` function which supports wildcards. So if a Moleculer developer wants to use it, he/she expect it works the same way. 
3. Why is the span excluded in the `spanFinish`? If some exporter processes the span in `spanStart`, it expects that the `spanFinish` will be called with the same span. If not, it can cause a memory leak in the exporters. By the way, would be better to implement the logic to the base exporter (similar to metrics reporters). In this case, you can do that, only filtering the spans in Jaeger/Zipkin, but not filtering in Console/Event exporters. More opportunities with the same function.
4. missing unit tests.

So sorry, but I close it in this form.

Issue #1014, opened 2021-09-24
Add support to define services in ESM modules, or using ESM-only dependencies


piranna:
At https://github.com/moleculerjs/moleculer/issues/328 there was previous work to add support for ESM modules, but as shown in https://github.com/moleculerjs/moleculer/issues/328#issuecomment-924311630, it's not working. Since `moleculer` is using `require()` everywhere, once that a service use an ESM-only dependency or the service itself is in a ESM module file, running `moleculer-runner` crash on boot. (`moleculer-runner` is the first one [using `require()`](https://github.com/moleculerjs/moleculer/blob/2ae6fa7ccd21682f8046d6b69954712ebc46fc41/bin/moleculer-runner.js#L10) itself...)

To fix that we have two alternatives:

1. The first (and ideal) one is to update `moleculer` source code to use ESM modules itself instead of CommonJS. All maintained versions of Node.js already has unflagged support for ESM modules, so there would not be any compatibility issues, but would be a bit of work and probably it would be set for another major version (the `v0.16`, in this case).

2. The quick solution would be to make use of the [esm](https://www.npmjs.com/package/esm) package, that provides a replacement of the `require()` function that has support for ESM modules, and use it at https://github.com/moleculerjs/moleculer/blob/2ae6fa7ccd21682f8046d6b69954712ebc46fc41/src/service-broker.js#L773 instead of the Node.js builtin one, so ESM services or their ESM dependencies can be loaded without needing to do any other change in the current code base of `moleculer`. If you are interested, we can provide a PR with this change, meanwhile you work towards the first option.

icebob:
I know that nowadays many OS maintainers have been starting to change all their packages from CJS to ESM and generate a lot of extra works for other dependent library maintainers. But I don't plan to start it with Moleculer packages and make issues for developers who use Moleculer in their projects. So option #1 is not an alternative (at least for now, maybe in later major version).

Option #2: I've checked the `esm` package but it looks it's not maintained, last version is published in 2019 however there are reported unresolved issues.

piranna:
Why got this clossed?

icebob:
Because PR created

piranna:
Ok, didn't see it before, thanks :-D

PR #1063, opened 2022-02-19
ESM Runner

icebob:
## :memo: Description

This PR implements an ESM compatible Moleculer Runner.
The Runner is able to load ESM configuration (`moleculer.config.mjs`) and ESM services.
```bash
moleculer-runner-esm --repl services/**/*.service.mjs
```

The index file wrapped into `index.mjs` so you can import internal modules. E.g:
```js
import { ServiceBroker, Errors } from "moleculer";
```

**Hot-reload doesn't work with ESM Runner:** https://github.com/nodejs/modules/issues/307
Node maintainers try to solve the missing features (module cache and module dependency tree) with [loaders](https://nodejs.org/api/esm.html#loaders) but this API is not stable yet.

### :dart: Relevant issues
#1014

### :gem: Type of change

### :scroll: Example service file `greeter.service.mjs`
```js
export default {
    name: "greeter",

    actions: {

        hello: {
            rest: {
                method: "GET",
                path: "/hello"
            },
            async handler() {
                return "Hello Moleculer";
            }
        }
    }
};
``` 

### :scroll: Example config file `moleculer.config.mjs`
```js
export default {
    namespace: "esm",
    nodeID: null
};
```

- [ ] create ESM-based project template `project-esm`
- [x] create an ESM-based test project and run as typescript test project


icebob:
You can test it, if you install `npm i moleculerjs/moleculer#runner-esm`

PR #1064, opened 2022-02-23
[TS] Update types for REPL

AndreMaz:
This PR replaces Vorpal-related types with the ones provided by native REPL

Context: https://github.com/moleculerjs/moleculer-repl/releases/tag/v0.7.0

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -992,6 +992,17 @@ declare namespace Moleculer {
 		version?: string|number;
 	}
 
+	interface MoleculerRepl extends REPLServer {
+		context: ReplContext
+	}
+
+	interface ReplContext extends VMContext {
+		/**
+		 * Moleculer Service Broker instance
+		 */
+		broker: ServiceBroker
+	}
+
 	namespace Loggers {
 		class Base {
 			constructor(opts?: GenericObject);
@@ -1042,7 +1053,7 @@ declare namespace Moleculer {
 		start(): Promise<void>;
 		stop(): Promise<void>;
 
-		repl(): REPLServer;
+		repl(): MoleculerRepl;
 
 		errorHandler(err: Error, info: GenericObject): void;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,5 @@
 import { EventEmitter2 } from "eventemitter2";
+import { REPLServer } from "repl";
 
 declare namespace Moleculer {
 	/**
@@ -991,12 +992,6 @@ declare namespace Moleculer {
 		version?: string|number;
 	}
 
-	interface MoleculerRepl extends Vorpal{
-		removeIfExist(command:string): void;
-	}
-
-
-
 	namespace Loggers {
 		class Base {
 			constructor(opts?: GenericObject);
@@ -1047,7 +1042,7 @@ declare namespace Moleculer {
 		start(): Promise<void>;
 		stop(): Promise<void>;
 
-		repl(): MoleculerRepl;
+		repl(): REPLServer;
 
 		errorHandler(err: Error, info: GenericObject): void;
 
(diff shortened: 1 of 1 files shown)


shawnmcknight:
I've been giving some thought recently to how best to handle types that are introduced to moleculer via a plugin such as the repl.  That is, these types are not naturally available in the moleculer core library but require, in this case, the installation of `moleculer-repl`.  I think the best approach here is to use TypeScript's [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) functionality.  This should allow types to be defined in `moleculer-repl` but augment the moleculer core types.  That is, if a consumer installs just `moleculer` then they wouldn't have the repl's types available unless the repl was also installed.

Here is an example of a repo where I augmented moleculer core's types: https://github.com/shawnmcknight/moleculer-graphql/blob/main/src/index.ts.  In that, the `ActionSchema` type was augmented to add the additional `graphql` property to it.  Consumers can get type safety without having to adjust moleculer core's types:
image

So the overall idea here is that I think all of the repl types should be removed from moleculer core and relocated to an `index.d.ts` file in `moleculer-repl`.   The `BrokerOptions` interface should be augmented to add `replCommands` and `replDelimiter` and the `ServiceBroker` class should be augmented to add the `repl` method.

Alternatively, we could continue the existing pattern.  Let me know your thoughts or if you have any questions.  Thanks.

icebob:
@shawnmcknight I like your idea, because with this feature, we can extend the Service schema from modules as well (e.g. database or channels module extends with properties the service schema type).

@AndreMaz could you make this modification? Or do you need Shawn's help?

AndreMaz:
@icebob yep, will look into it 😉

AndreMaz:
Hi @shawnmcknight I've generated types from JSdocs and added this block over here https://github.com/moleculerjs/moleculer-repl/blob/d11aefb682efb95c10a9e8dee745432a78d79612/types/src/index.d.ts#L71-L86

After doing it and after removing all REPL-related types from moleculer's `index.d.ts` things seem to work. 
image

However, I'm a complete noob when it comes to TS so I would appreciate if you could take a look at it.

shawnmcknight:
> However, I'm a complete noob when it comes to TS so I would appreciate if you could take a look at it.

@AndreMaz Okay so I took a look through this.  I'm not sure what your branch for `moleculer` currently looks like, but in doing some testing I basically removed everything that had "repl" in its name.  Then a few quick tweaks to your types in `moleculer-repl`:

```ts
declare module "moleculer" {
	interface ServiceBroker {
		repl(opts?: Partial<REPLOptions>): nodeRepl.REPLServer;
	}

	interface BrokerOptions {
		/**
		 * Custom command definition
		 */
		replCommands?: Array<CustomCommand>;
		/**
		 * REPL delimiter
		 */
		replDelimiter?: string;
	}
}
```

So the two changes I made were:

1.  Changed `replDelimiter` from `String` to `string`.  That may be just a typo, but in case, here is a SO link explaining the difference: https://stackoverflow.com/a/14727461/961933.
2. I removed the `MoleculerREPL` class and instead created an interface for `ServiceBroker`.  This should have the effect of augmenting the existing moleculer `ServiceBroker` class.  Since in my testing I removed the `repl()` method from the `ServiceBroker` definition in `moleculer`, this basically adds it back.  After that I can see:
image

I think that should do the trick.

AndreMaz:
>  but in doing some testing I basically removed everything that had "repl" in its name

Yeah, I've done same thing.

Thank you for your help 👍

index.d.ts
```diff
@@ -1,4 +1,5 @@
 import { EventEmitter2 } from "eventemitter2";
+import { Context as VMContext } from 'vm'
```
shawnmcknight, line 2: I think this can be dropped now.

AndreMaz: Ups! Removed

icebob:
@shawnmcknight can we merge it?

shawnmcknight:
@icebob @AndreMaz I think this is good to go as long as the REPL PR is updated at the same time.

This is one of those weird "it's actually a breaking change, but just to the types, and only to those who happen to use the affected types" scenarios though.  I think its fine to merge as long as long as a new version of the REPL is published at the same time and the changelog indicates that you should also get the latest of that.

PR #1068, opened 2022-03-12
fix missing or incorrect typings

fugufish:
## :memo: Description

fixes small bug in typing in a few typings

### :gem: Type of change

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -749,7 +749,7 @@ declare namespace Moleculer {
 		static mergeSchemaUnknown(src: GenericObject, target: GenericObject): GenericObject;
 	}
 
-	type CheckRetryable = (err: Error) => boolean;
+	type CheckRetryable = (err: MoleculerError) => boolean;
 
 	interface BrokerCircuitBreakerOptions {
 		enabled?: boolean,
--- a/index.d.ts
+++ b/index.d.ts
@@ -810,16 +810,6 @@ declare namespace Moleculer {
 		options?: GenericObject
 	}
 
-	interface ReplCommand {
-		command: string;
-		description: string;
-		alias?: string;
-		options: Array<object>,
-		types: object,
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -810,6 +810,16 @@ declare namespace Moleculer {
 		options?: GenericObject
 	}
 
+	interface ReplCommand {
```
shawnmcknight, line 813: Repl options have been migrated to the types files in `moleculer-repl` via moleculerjs/moleculer-repl#62.  If changes are necessary to the repl types, please create a PR in that repo instead of this one.

index.d.ts
```diff
@@ -878,6 +888,8 @@ declare namespace Moleculer {
 		skipProcessEventRegistration?: boolean;
 
 		maxSafeObjectSize?: number;
+
+		replCommands?: Array<ReplCommand>;
```
shawnmcknight, line 892: Similar note here regarding repl types going in the `moleculer-repl` repo.

index.d.ts
```diff
@@ -749,7 +749,7 @@ declare namespace Moleculer {
 		static mergeSchemaUnknown(src: GenericObject, target: GenericObject): GenericObject;
 	}
 
-	type CheckRetryable = (err: Error) => boolean;
+	type CheckRetryable = (err: Errors.MoleculerError) => boolean;
```
shawnmcknight, line 752: I don't necessarily see a problem with this, but I'm not 100% certain that all errors here are instances of `MoleculerError`.  @icebob Can you confirm for me if all errors passed to the retry check callback are always instances of `MoleculerError` or can they be other errors as well?

shawnmcknight (changes requested):

index.d.ts
icebob: No, it will be called with any Error that services can throw.

PR #1073, opened 2022-03-30
d.ts: allow custom ctx.locals type

0x0a0d:


index.d.ts
```diff
@@ -521,7 +521,7 @@ declare namespace Moleculer {
 		disconnected(): void;
 	}
 
-	class Context<P = unknown, M extends object = {}> {
+	class Context<P = unknown, M extends object = {}, L extends object = GenericObject> {
```
shawnmcknight, line 524: I don't believe there are any constraints to what type `ctx.locals` can be.  According to the docs, this is of `any` type.  I don't think this `extends` is necessary.

In inspecting the moleculer source, it appears that this property is initialized to an empty object.  I'd lean towards the default value of this type being `Record<string, never>` since that's the actual type in place unless the consumer makes changes to it.  So, I think this should be `L = Record<string, never>`.

0x0a0d: I think `L = GenericObject` will better, It will keep old code work without changes

Screen Shot 2022-03-31 at 03 07 10

shawnmcknight (approved):
Looks good.  I'm okay with using `GenericObject` for backwards compatibility.

icebob:
Thank you both!

PR #1074, opened 2022-03-31
Split load service

0x0a0d:
add method `loadServiceSchema` that allow to load a `ServiceSchema`
Usage on middleware createService


icebob:
Please explain why it's better, and show a use-case.

0x0a0d:
It can be used to create a service from schema instead path of file. Maybe only me need this so I close this pull

Issue #1072, opened 2022-03-29
Ctx is undefined when using shard strategy and preferLocal is false, throws error on emit


jscottsf:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Current Behavior

Balanced events via broker.emit with two nodes, shard strategy, and preferLocal false throws the error `Cannot read property 'params' of undefined`. Basically, the line `return _.get(ctx.params, this.opts.shardKey);` in the shard strategy function `getKeyFromContext` throws an error since the context is undefined and `params` cannot be accessed.

## Expected Behavior

A balanced event that is not-local should work. This can be fixed by:

1. Modify `getBalancedEndpoints` in event-catalog to accept `ctx` as the last argument.
2. Modify `getBalancedEndpoints` to pass the context in the `next` call, e.g. `const ep = list.next(ctx);`
3. In the `emit` function in service-broker, pass the built context as shown `getBalancedEndpoints(eventName, opts.groups, ctx);`

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Set preferLocal to false to ensure the balancer is used.
2. Enable the shard strategy which REQUIRES the context in order to access the shard key values.
3. Run two node instances.
4. Call broker.emit.

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.14.19
* NodeJS version: 14.17
* Operating System: Mac

icebob:
Yeah, please create a fix PR.

intech:
@jscottsf, please, check the error in my fork [`intech/moleculer`](https://github.com/intech/moleculer)

jscottsf:
Sorry that I didn't get to the PR yet. It's been on my list.

@intech Did a quick code review on the fork and it looks right. Noticed that emit became async and got some refactoring. Will this get pulled into the original repo? :)

intech:
@jscottsf It will but not now. The first is to deal with the performance slow due to refactoring.

intech:
@jscottsf сan I close this issue as resolved? I was not able to reproduce the problem during the tests.

jscottsf:
[...]

PR #1090, opened 2022-05-06
feat(strategies): async method select

intech:
@icebob It looks like there have a bug in the tracing integration test 

https://github.com/moleculerjs/moleculer/blob/master/test/integration/tracing.spec.js#L122

a function without await for the promise changing is breaking the test on the wrong snapshot 

https://github.com/moleculerjs/moleculer/blob/master/test/integration/tracing.spec.js#L189-L196 

Now I have not figured out this test to understand how to fix it.
I also need to fix it due to changes in event balancing for asynchronous calls.
The results of tracing in the snapshot will differ along with the change in the operation of events and calls.

Everything else works correctly. In benchmarks did not see any difference in speed.


intech:
I probably already fixed this bug https://github.com/moleculerjs/moleculer/issues/1072
There are fixes in the `next()` and `nextLocal()` functions where `ctx` was not passed.

intech:
We make benchmarks by methods to determine the cause of speed degradation.

intech:
Unfortunately, adding async/await reduces performance by half.

Issue #1065, opened 2022-02-24
Awaiting ctx.emit behave differently locally and remotely


Embraser01:
## :stop_sign: Do you want to ask a question ?

Please head to the [Discord chat](https://discord.gg/TSEcDRP)

## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Current Behavior

When awaiting a `ctx.emit` call, the promise wait for the event to be completed locally but will resolve as soon as the event is sent on the transit when sent remotely.

A workaround for now is to remove the await but if the `emit` is wrapped by a middleware, it prevents it to be async.

This can be found here: 

https://github.com/moleculerjs/moleculer/blob/1fa2440279ec8f242399e1f3dea30abead7c66e0/src/service-broker.js#L1357-L1389

## Expected Behavior

I would expect to have a consistent behavior for both cases. I'd say I'd expect the local event to not wait for completion (matching the remote behavior).

Note: It could break tests with events that relies on this behavior.

### Reproduce code snippet

```js
const { ServiceBroker } = require("moleculer");

const broker = new ServiceBroker();

broker.createService({
  name: "greeter",
  actions: {
    async hello(ctx) {
      const start = Date.now();
      await ctx.emit("my-event", { test: 1 });
      this.logger.warn(`Event emit took ${Date.now() - start}ms`);
    }
  },
  events: {
    "my-event": async () => {
      await new Promise((resolve) => setTimeout(resolve, 5000));
    }
  }
});

broker.start().then(() => {
  broker
    .call("greeter.hello", { name: "CodeSandbox" })
    .catch((err) => broker.logger.error(err.message));
});
```

Gives an output like this:
![image](https://user-images.githubusercontent.com/8802277/155505814-3333d3a0-7b5f-4432-8ca0-24009567a229.png)

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.14.19
* NodeJS version: LTS
* Operating System: Linux, Windows, MacOS

intech:
@icebob This one affects `emit` and `broadcast` with an enabled balancer, which becomes a problem.

PR #1096, opened 2022-06-05
fix(service-broker): local `emit` & `broadcast` handlers changed from async to event-driven

intech:
### :dart: Relevant issues
[issue#1065](https://github.com/moleculerjs/moleculer/issues/1065)

### :gem: Type of change


src/registry/event-catalog.js
```diff
@@ -35,6 +38,20 @@ class EventCatalog {
 		this.events = [];
 
 		this.EndpointFactory = EventEndpoint;
+
+		this.on("broker.event", ctx => {
```
icebob, line 42: Why we should you event emitter to call a non-await-ed method? I feel it too expensive. Not enough just skip the `await` in the `callEventHandler`

PR #1098, opened 2022-06-19
fix(service): removed static on methods for the use of ServiceFactory

intech:
## :memo: Description
To use custom property handlers inside services:
https://github.com/moleculerjs/moleculer/blob/452cbf33b438f3199329e0f4afe59429e4a08124/src/service.js#L630-L631
The problem is that a static method in the `Service` class is always called from that class (original).
Calling the `parseServiceSchema()` method from instance:
https://github.com/moleculerjs/moleculer/blob/452cbf33b438f3199329e0f4afe59429e4a08124/src/service.js#L57
causes the original static method from the `Service` class to be called
https://github.com/moleculerjs/moleculer/blob/452cbf33b438f3199329e0f4afe59429e4a08124/src/service.js#L74
It cannot be overridden via `ServiceFactory` and add custom handlers to combine properties.

### :gem: Type of change

@icebob @AndreMaz It is necessary to document the possibility of using custom handlers. I haven't seen any mention of this in the documentation anywhere.

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -145,8 +145,10 @@ class Service {
 						"stopped",
 						"_start",
 						"_stop",
-						"_init"
-					].indexOf(name) !== -1
+						"_init",
+						"applyMixins"
+					].indexOf(name) !== -1 ||
+					name.startsWith("mergeSchema")
 				) {
 					throw new ServiceSchemaError(
 						`Invalid method name '${name}' in '${this.name}' service!`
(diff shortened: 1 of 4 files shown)


src/service.js
```diff
@@ -145,7 +146,7 @@ class Service {
 						"_start",
 						"_stop",
 						"_init"
-					].indexOf(name) != -1
+					].indexOf(name) !== -1
 				) {
 					throw new ServiceSchemaError(
```
icebob, line 151: I've just known why these methods was static because they are not conflict with user-defined methods. 
To avoid it, please add a new check here and throw the same error, if the name  starts with `mergeSchema` or equals `applyMixins`

icebob (changes requested):

icebob (approved):
Good job, thanks!

PR #1099, opened 2022-06-22
Update event handler syntax to allow Promise<void> return value

shawnmcknight:
## :memo: Description

Currently, the types definition for event handlers specifies that the return value is `void`.  Internally, moleculer will allow for the processing of both synchronous and asynchronous functions as event handlers.  However, an async function returns `Promise<void>` which is not fully compatible with the existing definition.  This PR updates the types definition for `ServiceEventLegacyHandler` and `ServiceEventHandler` to allow for a function that either returns `void` or `Promise<void>`.

When saving, prettier automatically formatted the file to the current configuration in the repo.  Although that is not the intent of the PR, I figured it would be beneficial to get the formatting correct based on those standards.  Therefore, there are a lot of pure "formatting" changes in the PR as well.

### :gem: Type of change

Before this change, `@typescript-eslint` was reporting an error due to a promise returning function not being compatible with a void returning function ([rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-misused-promises.md)).  TypeScript itself allows functions with a `void` return to return `Promise<void>` but this rule exists to prevent a potential misuse.  I modified a local type definition of the moleculer types and neither `@typescript-eslint` nor typescript are reporting problems after the change.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -82,7 +82,7 @@ declare namespace Moleculer {
 			rate?: number | null;
 			tracesPerSecond?: number | null;
 			minPriority?: number | null;
-		}
+		};
 
 		actions?: boolean;
 		events?: boolean;
@@ -95,7 +95,7 @@ declare namespace Moleculer {
 		tags?: {
 			action?: TracingActionTags;
 			event?: TracingEventTags;
-		}
+		};
 	}
 
 	class Tracer {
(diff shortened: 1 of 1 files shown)


icebob (approved):
Thank you!

PR #1101, opened 2022-07-05
Fix info packet send on stop

Ubitso:
## :memo: Description

Use case is basic redeploy moleculer app with consistent event flow.
On redeploy we have large amount of ServiceIsNotAvailable errors.

Minimal reproduction repository:
https://github.com/Ubitso/moleculer-sd-issue

### :dart: Relevant issues
No

### :gem: Type of change

I've changed stop function in `service-broker`. Error was in regenerateLocalRawInfo in 'registry.js'. During broker stop process broker should send info packet with empty array of services. But it does send packet with current available services.
So we should properly send info packet, and then after delay set `broker.stopping = true`, otherwise remote services will send requests to registred services when it's locked by 'broker.stopping = true'

### :scroll: Example code
I've tried to reproduce using test but i cant get same errors. I think it happens only if services is in separate processes.
Please look at my test, if possible to reproduce in one process please give me advice to how achieve that.
```js

describe("Test proper broker stop", () => {
		transporter: "Redis",
		registry: {
			strategy: "RoundRobin",
			preferLocal: false,
			discoverer: {
				type: 'Local'
			}
		},

		tracking: {
			enabled: false
		}
	});

	jest.setTimeout(30000)

	it("Should not throw service is not available error", async () => {
		const errors = []

		broker.createService({
			name: "consumer",

			started() {
				const ctx = this.broker.ContextFactory.create(
					this.broker,
					null,
					{},
					{ caller: this.fullName }
				);

				let i = 0
				const interval = setInterval(async () => {
					if (i < 1000000000) {
						try {
							console.log('find called')
							await ctx.call("producer.find", {});
						} catch (err) {
							console.error(err)
							errors.push(err)
						}
					} else {
						console.warn('clean interval')
						clearInterval(interval)
					}
				}, 1);
			}
		});

		broker.createService({
			name: "producer",

			started() {},

			actions: {
				async find() {

					return new Promise((resolve) => {
						setTimeout(() => {
							return resolve({
								data: "some result"
							});
						}, 1);
					});
				},
			}
		});

		await broker.start();

		await broker.Promise.delay(2000)

		const brokerStopPromise = broker.stop();
		console.log('Broker stop called')

		await brokerStopPromise

		console.log(errors)
	});
});
``` 

I tested on my reproduction repo on versions 0.14.19 also on 0.14.21.
Also I fixed current test with fake timers because it starts to fail after my changes.

No

### DIFF
--- a/src/registry/registry.js
+++ b/src/registry/registry.js
@@ -443,7 +443,7 @@ class Registry {
 	 *
 	 * @memberof Registry
 	 */
-	regenerateLocalRawInfo(incSeq) {
+	regenerateLocalRawInfo(incSeq, isStopping) {
 		let node = this.nodes.localNode;
 		if (incSeq) node.seq++;
 
@@ -457,8 +457,12 @@ class Registry {
 			"seq",
 			"metadata"
 		]);
-		if (this.broker.started || incSeq) rawInfo.services = this.services.getLocalNodeServices();
-		else rawInfo.services = [];
+
+		if (!isStopping && (this.broker.started || incSeq)) {
+			rawInfo.services = this.services.getLocalNodeServices();
+		} else {
+			rawInfo.services = [];
+		}
 
 		// Make to be safety
 		node.rawInfo = utils.safetyObject(rawInfo, this.broker.options);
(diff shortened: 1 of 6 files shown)


icebob (approved):
Nice catch, thanks for your PR!

PR #1111, opened 2022-08-02
fix(typescript) updated check retryable error param type to moleculer error

alvarocantador:
## :memo: Description

Updated the CheckRetryable error param type to follow the real object structure and also what is described in the documentation:

```
...
check: err => err && !!err.retryable
...
```

The `retryable` property is only available in the `Errors.MoleculerError` type.

### :gem: Type of change

### :scroll: Example code
```js

``` 

- [ ] Test A
- [ ] Test B

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -804,7 +804,7 @@ declare namespace Moleculer {
 		static mergeSchemaUnknown(src: GenericObject, target: GenericObject): GenericObject;
 	}
 
-	type CheckRetryable = (err: Error) => boolean;
+	type CheckRetryable = (err: Errors.MoleculerError) => boolean;
 
 	interface BrokerCircuitBreakerOptions {
 		enabled?: boolean;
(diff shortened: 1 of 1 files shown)


shawnmcknight:
Please see this [comment](https://github.com/moleculerjs/moleculer/pull/1068#discussion_r828423923) from an earlier declined PR.  The retry middleware has to be capable of handling any error type as there is no guarantee that the error will be an instance of `Moleculer.Error`.

Since an attempt to change this has been submitted 2x now, I'm assuming people are having an issue making their own implementation of this check with the existing typings.  I did a quick test and with the type being `Error` alone it's not able to access the `retryable` property in an implementation.  I think the correct solution here is therefore:

```
type CheckRetryable = (err: Errors.MoleculerError | Error) => boolean;
```

An implementation can then narrow the type to access the retryable property without getting type errors:
```
const retry: CheckRetryable = (err) => {
    return err != null && 'retryable' in err && err.retryable;
}
```

shawnmcknight (changes requested):
I think the correct solution here has to be a union of `Errors.MoleculerError` and `Error`.

alvarocantador:
@shawnmcknight I agree with the type union, and not only the  `retryable` property but the `data` and `code` properties are also not available in the default `Error` type.

shawnmcknight (approved):
LGTM!

PR #1112, opened 2022-08-03
fixed the issue with setting tracing and metrics options with env variables

prasadmadanayake:
## :memo: Description
When using the runner we cannot set tracing options using env variable because defaultOption tracing is a boolean and runner only update the values if property is an object

https://github.com/moleculerjs/moleculer/blob/5d72c97b93896671f47352d11f9006c1080277f9/src/utils.js#L409

So the fix is created by changing the `tracing` and `metrics` property to object with default state as `false`

### :gem: Type of change

### :scroll: Example code
```shell
# now can set tracer type as below
MOL_TRACING__EXPORTER=Console 
# can set metrics type as below
MOL_METRICS__REPORTER=StatsD

``` 

Created a sample service schema and execueted using moleculer-runner manually.

### DIFF
--- a/src/service-broker.js
+++ b/src/service-broker.js
@@ -106,7 +106,7 @@ const defaultOptions = {
 	validator: true,
 
 	metrics: false,
-	tracing: false,
+	tracing: {enabled: false},
 
 	internalServices: true,
 	internalMiddlewares: true,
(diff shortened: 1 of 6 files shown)


icebob:
Could you make the same change with `metrics` as well, please?

prasadmadanayake:
> Could you make the same change with `metrics` as well, please?

updated

icebob (approved):
Thanks, good job!

Issue #1116, opened 2022-08-09
Configurable service dependencyTimeout in broker options


disce-omnes:
## Is your feature request related to a problem? Please describe.
Currently it's impossible to set [dependencyTimeout](https://moleculer.services/docs/0.14/services.html#Dependencies) in broker options, it's only possible on a per service level.

However there is already a configurable option for `dependencyInterval` both in [broker options](https://moleculer.services/docs/0.14/configuration.html#Broker-options) & in the service schema.

Configurable dependencyTimeout is useful, because with no timeout by default services indefinitely fail to start if their dependencies are unvailable. However it happens almost silently, since there is only [one log](https://github.com/moleculerjs/moleculer/blob/937d50d1ab6c03afdc688e3800cdb9cf5a95d49e/src/service-broker.js#L1027) for waiting for dependencies, which can be easily missed, while [additional logs](https://github.com/moleculerjs/moleculer/blob/937d50d1ab6c03afdc688e3800cdb9cf5a95d49e/src/service-broker.js#L1066) are on the debug level.

In a complex system with many nested dependencies it's difficult to find out which service is the root cause of why the system fails to start.

## Describe the solution you'd like
Similarly to `dependencyInterval` there could be `dependencyTimeout` option in broker options.

## Describe alternatives you've considered 
Global `dependencyTimeout` can be set in the base service class which concrete services inherit. However there might not be such a class or ES6 classes might not be used at all in legacy projects.

## Additional context
This change will not be breaking if default value for this option will be set to 0 (no timeout) same way it's currently [hard-coded](https://github.com/moleculerjs/moleculer/blob/937d50d1ab6c03afdc688e3800cdb9cf5a95d49e/src/service.js#L288).

I can submit a PR, if this change is approved.

PR #1118, opened 2022-08-10
add dependencyTimeout broker option

disce-omnes:
## :memo: Description

Add dependencyTimeout broker option with default 0 (no timeout). Currently there is no way to configure it in broker options, only on a per service level. 0 timeout is currently hard-coded, so there should be no breaking change.

See relevant issue for detailed justification.

### :dart: Relevant issues
https://github.com/moleculerjs/moleculer/issues/1116

### :gem: Type of change

### :scroll: Example code
```js
  dependencyTimeout: 60_000
});
``` 

I've added an additional test case for this option & ran all existing tests except for memory leak tests.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -913,6 +913,7 @@ declare namespace Moleculer {
 		internalMiddlewares?: boolean;
 
 		dependencyInterval?: number;
+		dependencyTimeout?: number;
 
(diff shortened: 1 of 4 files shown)


icebob:
Good job thx!

PR #1126, opened 2022-08-24
Allow moleculer-runner to resolve configuration files from node_modules

shawnmcknight:
## :memo: Description

Currently, `moleculer-runner` will resolve configuration files in the following ways:
1) If configuration options (environment variable or command line) set config to an absolute path, the configuration will be resolved from that absolute path
2) If configuration options (environment variable or command line) set config to a non-absolute path, the configuration will be resolved from a path relative to the current working directory
3) Resolved from a `moleculer.config.js` file in the current working directory
4) Resolved from a `moleculer.config.json` file in the current working directory

The current resolution strategy does not allow for resolution of configurations from `node_modules` because relative paths will only be loaded relative to the current working directory and will not follow the node resolution workflow to look at `node_modules`.  Moleculer users may wish to resolve from `node_modules` for the following reasons:
1) Configuration files are kept in separate external packages
2) Configuration files are kept in a different workspace in a monorepo.  Workspaces in a monorepo are symlinked through `node_modules` with `npm`, `yarn`, or `pnpm` workspaces.

This PR updates `moleculer-runner` so that it will resolve configuration files from `node_modules`.  After this change, the runner will resolve as follows:

**Environment Variable or Command Line Configuration Set**
1) Set to an absolute path then configuration will be resolved from that absolute path
2) Set to a non-absolute path then configuration will first be resolved from a path relative to the current working directory.
3) Set to a non-absolute path and resolution failed relative to the current working directory then resolution will be resolved using regular [node module resolution strategy](https://nodejs.org/api/modules.html#all-together) starting the resolution from the current working directory.

Effectively, item 3 is added to the resolution workflow.  Items 1 and 2 should be consistent with existing behaviors.  If resolution fails then the runner will abort.

**No Environment Variable or Command Line Configuration Set**
4) Resolved from a `moleculer.config.js` file in the current working directory
5) Resolved from a `moleculer.config.json` file in the current working directory

This is identical to the exiting workflow.  If no configuration exists then one will not be used.

### :gem: Type of change

There are no existing unit tests (that I could find) for the runner.  I tested in a local moleculer repository to ensure that the expected configurations were loaded.

1) Confirmed load from specified configuration file which was in the current working directory (`--config=test-config.js` and `--config=test.config.json`)
2) Confirmed load from specified configuration file which was in a subdirectory of the current working directory (`--config=./src/test-config.js` and `--config=./src/test.config.json`
3) Confirmed load for specified configuration file which was in a monorepo workspace (`--config=@scope/moleculer-config/test-config.js`
4) Confirmed load for `moleculer.config.js` from the current working directory with no specified configuration file
5) Confirmed load for `moleculer.config.json` from the current working directory with no specified configuration file

### DIFF
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -201,7 +201,7 @@ export default class MoleculerRunner {
 		for (let idx = 0; idx < paths.length; idx++) {
 			const path = paths[idx];
 			try {
-				return require.resolve(path);
+				return require.resolve(path, { paths: [process.cwd()] });
 			} catch (_) {
 				// ignore
 			}
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -155,24 +155,31 @@ export default class MoleculerRunner {
 		let filePath;
 		// Env vars have priority over the flags
 		const configPath = process.env["MOLECULER_CONFIG"] || this.flags.config;
-		if (configPath != null) {
-			const paths = path.isAbsolute(configPath)
-				? [configPath]
-				: [path.resolve(process.cwd(), configPath), configPath];
 
-			filePath = this.tryConfigPaths(paths);
+		if (configPath != null) {
+			if (path.isAbsolute(configPath)) {
+				filePath = this.tryConfigPath(configPath);
+			} else {
+				filePath = this.tryConfigPath(path.resolve(process.cwd(), configPath));
+
+				if (filePath == null) {
+					filePath = this.tryConfigPath(configPath, true);
+				}
+			}
 
 			if (filePath == null) {
 				return Promise.reject(new Error(`Config file not found: ${configPath}`));
 			}
-		} else {
-			filePath = this.tryConfigPaths([
-				path.resolve(process.cwd(), "moleculer.config.js"),
-				path.resolve(process.cwd(), "moleculer.config.json")
-			]);
 		}
 
-		if (filePath) {
+		if (filePath == null) {
+			filePath = this.tryConfigPath(path.resolve(process.cwd(), "moleculer.config.js"));
+		}
+		if (filePath == null) {
+			filePath = this.tryConfigPath(path.resolve(process.cwd(), "moleculer.config.json"));
+		}
+
+		if (filePath != null) {
 			const ext = path.extname(filePath);
 			switch (ext) {
 				case ".json":
@@ -193,21 +200,22 @@ export default class MoleculerRunner {
 	}
 
 	/**
-	 * Try to resolve a configuration file at a series of paths
-	 * @param {string[]} paths - Paths to attempt resolution from
+	 * Try to resolve a configuration file at a path
+	 * @param {string} configPath - Path to attempt resolution from
+	 * @param {boolean} [startFromCwd=false] - Start resolution from current working directory
 	 * @returns {string | null}
 	 */
-	tryConfigPaths(paths) {
-		for (let idx = 0; idx < paths.length; idx++) {
-			const path = paths[idx];
-			try {
-				return require.resolve(path, { paths: [process.cwd()] });
-			} catch (_) {
-				// ignore
-			}
+	tryConfigPath(configPath, startFromCwd = false) {
+		let resolveOptions;
+		if (startFromCwd) {
+			resolveOptions = { paths: [process.cwd()] };
 		}
 
-		return null;
+		try {
+			return require.resolve(configPath, resolveOptions);
+		} catch (_) {
+			return null;
+		}
 	}
 
 	normalizeEnvValue(value) {
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -173,6 +173,9 @@ export default class MoleculerRunner {
 		}
 
 		if (filePath == null) {
+			filePath = this.tryConfigPath(path.resolve(process.cwd(), "moleculer.config.mjs"));
+		}
+		if (filePath == null) {
 			filePath = this.tryConfigPath(path.resolve(process.cwd(), "moleculer.config.js"));
 		}
 		if (filePath == null) {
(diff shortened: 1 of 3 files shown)


src/runner.js
```diff
@@ -151,26 +151,32 @@ class MoleculerRunner {
...
-			filePath = path.isAbsolute(process.env["MOLECULER_CONFIG"])
-				? process.env["MOLECULER_CONFIG"]
-				: path.resolve(process.cwd(), process.env["MOLECULER_CONFIG"]);
-		} else if (this.flags.config) {
-			filePath = path.isAbsolute(this.flags.config)
-				? this.flags.config
-				: path.resolve(process.cwd(), this.flags.config);
+		const configPath = process.env["MOLECULER_CONFIG"] || this.flags.config;
+
+		if (configPath != null) {
```
intech, line 156: Please, use the strict equality operator (`===`)

src/runner.js
```diff
@@ -151,26 +151,32 @@ class MoleculerRunner {
...
+			if (path.isAbsolute(configPath)) {
+				filePath = this.tryConfigPath(configPath);
+			} else {
+				filePath = this.tryConfigPath(path.resolve(process.cwd(), configPath));
+
+				if (filePath == null) {
```
intech, line 162: Please, use the strict equality operator (`===`)

src/runner.js
```diff
@@ -151,26 +151,32 @@ class MoleculerRunner {
...
+					filePath = this.tryConfigPath(configPath, true);
+				}
+			}
+
+			if (filePath == null) {
```
intech, line 167: Please, use the strict equality operator (`===`)

src/runner.js
```diff
@@ -151,26 +151,32 @@ class MoleculerRunner {
...
+				return Promise.reject(new Error(`Config file not found: ${configPath}`));
+			}
 		}
-		if (!filePath && fs.existsSync(path.resolve(process.cwd(), "moleculer.config.js"))) {
-			filePath = path.resolve(process.cwd(), "moleculer.config.js");
+
+		if (filePath == null) {
```
intech, line 172: Please, use the strict equality operator (`===`)

src/runner.js
```diff
@@ -151,26 +151,32 @@ class MoleculerRunner {
...
+			filePath = this.tryConfigPath(path.resolve(process.cwd(), "moleculer.config.js"));
 		}
-		if (!filePath && fs.existsSync(path.resolve(process.cwd(), "moleculer.config.json"))) {
-			filePath = path.resolve(process.cwd(), "moleculer.config.json");
+		if (filePath == null) {
```
intech, line 175: Please, use the strict equality operator (`===`)

src/runner.js
```diff
@@ -151,26 +151,32 @@ class MoleculerRunner {
...
+			filePath = this.tryConfigPath(path.resolve(process.cwd(), "moleculer.config.json"));
 		}
 
-		if (filePath) {
-			if (!fs.existsSync(filePath))
-				return Promise.reject(new Error(`Config file not found: ${filePath}`));
-
+		if (filePath != null) {
```
intech, line 179: Please, use the strict equality operator (`===`)

intech (changes requested):

src/runner.js
shawnmcknight: I can make this change if that's the standard, but generally speaking my one exception to strict vs. loose equality checking is against `null` as loose equality will capture both `undefined` and `null` scenarios.  All other scenarios I would generally use strict equality instead, but this is done intentionally.  It looks like that behavior is used often in this repo (434 results for `!= null` and 63 results for `== null`).

There's actually an eslint mode for their strict equality rule to allow for this (https://eslint.org/docs/latest/rules/eqeqeq#smart) although it doesn't look like the "eqeqeq" rule is enabled for this repo.

As I said, I can make the change if that's the preference here.

intech: Thank you! If there can be `undefined` or `null`, then it's better to check it explicitly in the code.

PS: I would add a rule in eslint and check the entire repository for this in a separate PR. I consider this an important step to improve the quality of the code.

@icebob, what do you think?

intech: In fact, in a large amount of code with many such checks, debugging unexpected behaviour at the end can be an unpleasant surprise at one of these checks.

shawnmcknight: @intech I think it's a very good idea to turn on the `eqeqeq` eslint rule in the repo.  If the standard for the repo is to use strict equality its much better to catch that in lint than in PR review.

With that said, I'm personally a fan of allowing for the loose equality exception against `null`. In the vast majority of circumstances, `null` and `undefined` are treated equivalently and the only things `null` matches against loosely are itself and `undefined`.  I feel like its much more likely to introduce a bug for a value that could be either `null` or `undefined` and forgetting to check against both when using strict equality than it is to have a scenario where `null` and `undefined` are actually treated differently from each other and a bug is introduced that way.

Ultimately I'm good with either approach here, but my opinion is that loose equality against `null` is a good exception to the strict equality rule.

icebob: I don't like that forcing the conditions to `===`. As @shawnmcknight said, many places allow that `null` and `undefined` handled by the same way. When it's not true, I use the `===` instead of `==` in the codebases. And I check these situations in the PRs as well.

For me the `a == null` it's more readable than the `a === null || a === undefined`.

src/runner-esm.mjs
```diff
@@ -149,29 +154,32 @@ export default class MoleculerRunner {
...
+			}
+
+			if (filePath == null) {
+				return Promise.reject(new Error(`Config file not found: ${configPath}`));
+			}
 		}
-		if (!filePath && fs.existsSync(path.resolve(process.cwd(), "moleculer.config.js"))) {
-			filePath = path.resolve(process.cwd(), "moleculer.config.js");
+
+		if (filePath == null) {
```
icebob, line 175: The original `runner-esm` code tries to load `moleculer.config.mjs` as well.

icebob (changes requested):

src/runner-esm.mjs
shawnmcknight: Good catch. I've added a check for `moleculer.config.mjs` here.

icebob (approved):
Good job, thanks!

Issue #1129, opened 2022-08-29
Moleculer causes typescript errors with typescript version 4.8.2


SockeThe2nd:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository

## Current Behavior

The currents version of moleculer throws a typescript error when compiling with the current 4.8.2 version of typescript.

## Expected Behavior

The build should pass without errors

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. update typescript dependency of moleculer to 4.8.2  via `npm i typescript@4.8`
2. run the ts-test `npm run test:ts

The tests will fail. Trying to build a project with ts 4.8 also fail.

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.14.23 (current)
* Typescipt version: 4.8.2
* NodeJS version: 14.18.1
* Operating System: Ubuntu 20.01 under WSL2 in Win10

### Failure Logs
```
index.d.ts:757:3 - error TS2416: Property 'settings' in type 'Service<S>' is not assignable to the same property in base type 'ServiceSchema<ServiceSettingSchema>'.
  Type 'S' is not assignable to type 'ServiceSettingSchema | undefined'.
    Type 'S' is not assignable to type 'ServiceSettingSchema'.

757   settings: S;
      ~~~~~~~~

  index.d.ts:749:16
    749  class Service<S = ServiceSettingSchema> implements ServiceSchema {
                       ~~~~~~~~~~~~~~~~~~~~~~~~
    This type parameter might need an `extends ServiceSettingSchema` constraint.
  index.d.ts:749:16
    749  class Service<S = ServiceSettingSchema> implements ServiceSchema {
                       ~~~~~~~~~~~~~~~~~~~~~~~~
    This type parameter might need an `extends ServiceSettingSchema | undefined` constraint.


Found 1 error in index.d.ts:757
```

shawnmcknight:
I agree that having `S extends `ServiceSettingSchema` in all of these locations is a more correct interpretation of what the generic should be since the internal properties are present on the value of `settings` even if the consumer specified an interface that does not include them.

[...]

PR #1130, opened 2022-08-29
Fixes typesciprt-errors with typescript version 4.8.2

SockeThe2nd:
## :memo: Description

Fixes typescript-errors when using v4.8.2 of typescript

### :dart: Relevant issues

https://github.com/moleculerjs/moleculer/issues/1129

### :gem: Type of change

### :scroll: Example code

None needed. Current ts-tests `npm run test:ts` fail when using v4.8 of typescript

Updated the typescript-version in the project to 4.8.2, then ran the ts-test `npm run test:ts`

  - Old tests failed with current version of ts

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -166,18 +166,18 @@ declare namespace Moleculer {
 	type TracingActionTags =
 		| TracingActionTagsFuncType
 		| {
-				params?: boolean | string[];
-				meta?: boolean | string[];
-				response?: boolean | string[];
-		  };
+			params?: boolean | string[];
+			meta?: boolean | string[];
+			response?: boolean | string[];
+		};
 
 	type TracingEventTagsFuncType = (ctx: Context) => GenericObject;
 	type TracingEventTags =
 		| TracingEventTagsFuncType
 		| {
-				params?: boolean | string[];
-				meta?: boolean | string[];
-		  };
+			params?: boolean | string[];
+			meta?: boolean | string[];
+		};
 
 	interface TracingOptions {
 		enabled?: boolean;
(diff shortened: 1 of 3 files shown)


shawnmcknight:
I thought I had commented on this the other day, but I'm not sure what happened.

While I agree with the change (i.e. all of these locations _should_ extend `ServiceSettingSchema`) this would represent a breaking change for any consumers that do not currently have a settings type that does extend `ServiceSettingSchema`.  Those consumers could be using a type which does not without causing any problem provided they did not attempt to access any of the built-in properties on `settings`.  So, this change as is I believe would represent a breaking change for existing users and would need to be deferred until a new release.

In the meantime, I am wondering if instead of this breaking change if changing:

```
class Service<S = ServiceSettingSchema> implements ServiceSchema {
```
to
```
class Service<S = ServiceSettingSchema> implements ServiceSchema<S> {
```
would bypass the TS errors and also avoid the breaking change.  If that change works, it could be merged immediately to resolve the TS errors and then allow this proposed change to be deferred until a breaking change release was ready to go.

icebob (changes requested):
Please follow @shawnmcknight change requests

SockeThe2nd:
I implemented the requested change for the Service class. I am unsure if further changes to the internal types/interfaces I touched are needed.

icebob:
@shawnmcknight could you approve it, plz?

index.d.ts
```diff
@@ -475,7 +475,7 @@ declare namespace Moleculer {
 		basePath?: string;
 	}
 
-	type ActionSchema<S = ServiceSettingSchema> = {
+	type ActionSchema<S extends ServiceSettingSchema = ServiceSettingSchema> = {
```
shawnmcknight, line 478: My concern regarding the breaking change is the use of `extends ServiceSettingSchema` in all of these places.  That will break any consumers that are not currently extending that interface.  Can you try eliminating all of the `extends ServiceSettingSchema` from all locations and see if the other change to the implements clause in the `Service` class resolves the overall errors with TS?

shawnmcknight (changes requested):

Issue #1138, opened 2022-09-21
Typescript definitions don't match the actual Javascript implementation


SKarajic:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Current Behavior
The current version of the Typescript definitions don't match the actual Javascript implementation, such as how mergeSchemas and applyMixins are no longer static classes.

## Expected Behavior
To be able to import the code in Typescript and have them refer to the proper Javascript implementation.

## Failure Information
Multiple methods do not align with the Javascript implementation. If you were to go by the types, your code would not run as the implementation doesn't match. Calling the proper implementation will also upset Typescript as it does not have those definitions.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. use a method such as mergeSchemas which is a static method according to typescript.
2. run your code.

The process will fail as mergeSchemas does not exist as a static method anymore.

### Reproduce code snippet

```ts
import {Service, ServiceBroker, ServiceSchema} from "moleculer";

export default class SomeService extends Service {
  /*...*/
  // @ts-ignore
  public  constructor(public broker: ServiceBroker, schema: ServiceSchema<{}> = { name: '...' }) {
    super(broker);
    this.parseServiceSchema(Service.mergeSchemas({ /*...*/ })
  }
  /*...*/
}
```

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.14.23
* NodeJS version: 16.16.0
* Operating System: Linux

### Failure Logs
```
[Runner] moleculer_1.Service.mergeSchemas is not a function TypeError: moleculer_1.Service.mergeSchemas is not a function
```

saeedtabrizi:
I had same issue , but I fixed this issue  via removing all `mergeSchemas` methods  from service classes .

SKarajic:
In case anyone wants to patch their typescript file, I've temporarily made a tag on my fork with the fixes ([link over here](https://github.com/skarajic/moleculer/tree/0.14.23-patched)). To use the patch, replace your dependency version in the package.json with `https://github.com/skarajic/moleculer#0.14.23-patched`. I'll be deleting the tag again once my merge request is merged into the repository and a new release has been made 😃

Issue #1133, opened 2022-08-31
mergeSchemas is no longer available as a static class method of Service


ostracod:
## :stop_sign: Do you want to ask a question ?

Please head to the [Discord chat](https://discord.gg/TSEcDRP)

## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository

## Current Behavior

Under Moleculer 0.14.23, the static class method `mergeSchemas` of `Service` has been changed to be an instance method. As a result, `Service.mergeSchemas` is `undefined`.

## Expected Behavior

I expect `mergeSchemas` to be available as a static class method of `Service`, as documented here: https://moleculer.services/docs/0.14/api/service.html#mergeSchemas

## Failure Information

This breaks compatibility with our microservices which rely on `Service.mergeSchemas`.

### Steps to Reproduce

1. Create a project which has Moleculer 0.14.23 as the only dependency.
2. Run the code snippet below.
3. The output is `undefined`, but I am expecting `[Function: mergeSchemas]`. 

### Reproduce code snippet

```js
const { Service } = require('moleculer');
console.log(Service.mergeSchemas);
```

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: 0.14.23
* NodeJS version: 16.16.0
* Operating System: macOS 12.5.1

### Proposed Resolution

We would appreciate if you did one of the following:

- Expose `Service.mergeSchemas` again as specified by the documentation.

OR:

- Update the documentation to specify that `mergeSchemas` is an instance method. It is not a huge deal to call `this.mergeSchemas`, but we want to make sure our code won't break again because of a compatibility issue.

Thanks for any help you can give on this issue.

intech:
@ostracod Thanks for your feedback!
As I understood from the proposed solution, updating the documentation would be a good compromise.
You can read about [this issue](https://github.com/moleculerjs/moleculer/issues/1124#issuecomment-1218366223).

ostracod:
@intech Thanks for updating the docs!

PR #1139, opened 2022-09-21
Fix typescript definitions for the Service class

SKarajic:
## :memo: Description
The current version of the Typescript definitions don't match the actual Javascript implementation, such as how mergeSchemas and applyMixins are no longer static classes.

### :dart: Relevant issues
#1138 
#1133 

### :gem: Type of change

- [x] Tested using the `npm run test:ts` command

  - There is no framework set up to write tests for the index.d.ts file


disce-omnes:
I might be wrong, but can't tests in `test/typescript/tsd` be used to test index.d.ts?

index.d.ts
```diff
@@ -746,11 +746,23 @@ declare namespace Moleculer {
 		statuses: Array<{ name: string; available: boolean }>;
 	}
 
-	class Service<S = ServiceSettingSchema> implements ServiceSchema {
-		constructor(broker: ServiceBroker, schema?: ServiceSchema<S>);
-
+	class Service<S = ServiceSettingSchema, SM = ServiceSettingSchema> implements ServiceSchema {
```
icebob, line 749: Could you explain why `SM` needs?

SKarajic: In case you want to have a certain definition of the schema mod settings, you can strictly type the schema mods. Is it very useful? Not really, but i do believe that people should have at least the ability to enforce the schema mod structure.

icebob: @shawnmcknight what do you think?

shawnmcknight: The main reason for the `S = ServiceSettingSchema` generic was to allow `this.settings` to be typed so it wasn't basically `Record<string, any>`.  It wasn't really to gain any sort of type safety in the `Service` constructor, although it would appear that ended up as a side effect since the `schema` parameter is typed.  I would suggest that if you're trying to merge schemas with disparate `settings` properties via providing `schemaMods` then this is going to turn out badly because you either won't get the proper type of the `schema` parameter or the wrong type for the `settings` property since they use the same generic.

Is there any external use case for `schemaMods` being provided to the constructor?  That is, for what reason would an external consumer provide this parameter to the `Service` constructor.  I do see it documented in the `ServiceBroker` API and it looks like it was recently added when the static members of `Service` were made instance members.  What is the use case for this?  Is there one for which it makes sense to expose this as a parameter for external consumption?

icebob: @intech could you explain the new `schemaMods` parameter of service constructor?

SKarajic: > The main reason for the `S = ServiceSettingSchema` generic was to allow `this.settings` to be typed so it wasn't basically `Record<string, any>`. It wasn't really to gain any sort of type safety in the `Service` constructor, although it would appear that ended up as a side effect since the `schema` parameter is typed. I would suggest that if you're trying to merge schemas with disparate `settings` properties via providing `schemaMods` then this is going to turn out badly because you either won't get the proper type of the `schema` parameter or the wrong type for the `settings` property since they use the same generic.
> 
> Is there any external use case for `schemaMods` being provided to the constructor? That is, for what reason would an external consumer provide this parameter to the `Service` constructor. I do see it documented in the `ServiceBroker` API and it looks like it was recently added when the static members of `Service` were made instance members. What is the use case for this? Is there one for which it makes sense to expose this as a parameter for external consumption?

Seems like the main use case for it seems to be testing. The typescript starter includes some tests where schemaMods is used in order to override an action so it could be checked using jest. As for more practical use cases... not sure. Changing it would be a breaking change though.

intech: @icebob It has been ported from [service-broker](https://github.com/moleculerjs/moleculer/pull/1098/files#diff-3cd4f90d2dcbb1220a816dee13337a1f0d6853c2da7b5bf61bf10c68ce62e43aL827-L830) to [service](https://github.com/moleculerjs/moleculer/pull/1098/files#diff-145d31b275fbdb6a7b17ef982f83babd2323896502f2c8374da313a8833e0ebaR51-R57) to preserve the ability to call the `mergeSchemas` method inside the instance.
I seem to have seen its use only in tests.

SKarajic: Not to be annoying, but if this issue is not being discussed, and there isn't any other feedback provided, could we look into getting the MR merged?

Not having these types fixed means typescript is broken on this and upcoming versions as long for anyone who is using any of the schema merge methods.

AndreMaz:
Hi @shawnmcknight what do you think about it? Can it be merged?

index.d.ts
icebob: So, I've checked the source and figured out the root of the `schemaMods`. It's a very old "feature", it was implemented in early versions [6 years ago](https://github.com/moleculerjs/moleculer/commit/6628763e93bcf86fb650bca3b31caf2820db619e) when the `mixins` logic is not implemented yet. So this parameter was added to extend an existing service schema. 
But since the mixins feature, it's not used, but we cannot remove it because it can cause breaking change if somebody still uses it. So I removed it from the d.ts, but kept it in the code and added a deprecation warning. So we will able to remove it in version 0.15.

Moreover, I've removed from the PR the extra exposed internal methods. I don't like if we expose the internal methods because the users can use them directly because it will cause problems when we refactor them.

index.d.ts
```diff
@@ -763,12 +763,20 @@ declare namespace Moleculer {
 		logger: LoggerInstance;
 		actions: ServiceActions;
 		Promise: PromiseConstructorLike;
-		//currentContext: Context | null;
 
 		_init(): void;
```
shawnmcknight, line 767: With eliminating the "internal" methods, I believe these should be removed as well.

index.d.ts
```diff
@@ -779,29 +787,118 @@ declare namespace Moleculer {
 		 * @param logger the broker logger instance
```
shawnmcknight, line 787: The `logger` parameter was eliminated but the jsdoc was left behind.  It looks like there is no `logger` parameter so the jsdoc should be removed.

shawnmcknight:
Looked through post @icebob 's changes.  There were just two minor notes I added but otherwise I think this is good to go.

index.d.ts
icebob: I kept it because it was eliminated early, and can cause breaking if I remove it and somebody uses them. So I removed that is added in this PR only.

icebob:
Thanks @shawnmcknight

PR #1145, opened 2022-10-18
Typescript - add type to this in hooks

someone635:
## :memo: Description

In typescript, you can can typing of this by turning on noImplicitThis: true in tsconfig. Everything was working fine in my project until I used a service hook, where typescript showed the error:

'this' implicitly has type 'any' because it does not have a type annotation.ts(2683)
users.service.ts(77, 20): An outer value of 'this' is shadowed by this container.

I investigated and found that elsewhere in the library, this was typed using ThisType utility. I tried applying that to hooks, but it did not work, as hooks can be in arrays, whereas ThisType can only be applied to objects with methods (as far as I understand).

So I overwrote the this type directly in ActionHook types and it worked.

🎯 Relevant issues
See PR https://github.com/moleculerjs/moleculer/pull/980 . A debate occurred over using ThisType over this override, as this override caused issues in the rest of the library. I don't have enough experience with the library to know if using this override for hooks causes other issues, but I believe hooks are not used as globally as say, action or method handlers, so I don't think this change will have big repercussions. 

### :gem: Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

I used the Typescript template [Typescript template](https://github.com/moleculerjs/moleculer-template-project-typescript), to which I added NoImplicitAny:true in tsconfig. Then I added service hooks in arrays, as the template does not use any. Errors appeared, and the change made the errors disappear.

- I haven't tested what "this" is supposed to be in an Action Hook (aka not Service Hook). I suppose "this" in Action Hooks is the same as in Service Hooks, but I have no confirmation from the docs, and I can't really test it because of other pending Typescript bug: https://github.com/moleculerjs/moleculer/issues/1138. This change will affect both types of hooks, so if someone has more experience with the library, please confirm.

- [X ] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [X ] **I have added tests that prove my fix is effective or that my feature works**

EDIT: Did the same change on created, started, stopped lifecycle hooks

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -725,9 +725,9 @@ declare namespace Moleculer {
 		hooks?: ServiceHooks;
 
 		events?: ServiceEvents;
-		created?: (() => void) | Array<() => void>;
-		started?: (() => Promise<void> | void) | Array<() => Promise<void> | void>;
-		stopped?: (() => Promise<void> | void) | Array<() => Promise<void> | void>;
+		created?: ((this: Service<S>) => void) | Array<(this: Service<S>) => void>;
+		started?: ((this: Service<S>) => Promise<void> | void) | Array<(this: Service<S>) => Promise<void> | void>;
+		stopped?: ((this: Service<S>) => Promise<void> | void) | Array<(this: Service<S>) => Promise<void> | void>;
 
 		[name: string]: any;
 	}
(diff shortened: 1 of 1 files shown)


someone635:
Closing this as I'm preparing something better

PR #1149, opened 2022-11-02
Proposal - Allow matching hooks to multiple actions with "|"

someone635:
## :memo: Description

For example, you could write "create|update" to match a hook to both the create and update actions. A single edit to a line of code was required to add this functionality, as a regex matcher is already present in the code. This simply tells the matcher to use the regex matcher if a | is detected in the hook name. This is only breaking if for some reason people used | already in their action names.

### :dart: Relevant issues
No issue that I know of

### :gem: Type of change

- [x] New feature (non-breaking change which adds functionality) (only very minor breaking)

- [x] Added simple unit tests
- [x] Using the patch via patch-package in my own code and it works.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -306,7 +306,7 @@ const utils = {
 	 */
 	match(text, pattern) {
 		// Simple patterns
-		if (pattern.indexOf("?") == -1 && pattern.indexOf("|") == -1) {
+		if (pattern.indexOf("?") == -1) {
 			// Exact match (eg. "prefix.event")
 			const firstStarPosition = pattern.indexOf("*");
 			if (firstStarPosition == -1) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -306,7 +306,7 @@ const utils = {
 	 */
 	match(text, pattern) {
 		// Simple patterns
-		if (pattern.indexOf("?") == -1) {
+		if (pattern.indexOf("?") == -1 && pattern.indexOf("|") == -1) {
 			// Exact match (eg. "prefix.event")
 			const firstStarPosition = pattern.indexOf("*");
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -306,7 +306,7 @@ const utils = {
 	 */
 	match(text, pattern) {
 		// Simple patterns
-		if (pattern.indexOf("?") == -1) {
+		if (pattern.indexOf("?") == -1 && pattern.indexOf("|") == -1) {
```
icebob, line 309: You can't modify the logic of `match` method in utils because it is used in many places in the core and official modules, to match service names, action names, event names...etc. So this change affects other search logics as well and it can cause major breaking changes, too.

icebob (changes requested):

src/utils.js
someone635: Oh I see, I thought it was only used by hooks. You wouldn't want to allow users to add "a|b" as an action, service or event name, which would have undefined behavior...

But just to know, do you think the feature proposal by itself is a good addition to the framework?

icebob: Yeah, this change can be a good feature, but the pipe sign should be parsed inside the ActionHooks middleware only.

someone635: Great! I'll get to it sometimes this week.

icebob (approved):
Good job, thanks!

icebob:
I've just released it in v0.14.26. Could you help add it to the [Moleculer documentation](https://github.com/moleculerjs/site/blob/master/source/docs/0.14/actions.md#action-hooks), as well?

someone635:
Sure thing, sorry, forgot about it...

PR #1161, opened 2022-12-03
feat(types): add middlewares types

thib3113:
Add typing for the optionals middlewares

PS : I typed the encryption middleware too, but I think it's not a good middleware, not secure (IV not changing), and hard to use (password length need to be a strict length)

### :gem: Type of change

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,5 @@
 import { EventEmitter2 } from "eventemitter2";
+import type { Kleur } from "kleur";
 
 declare namespace Moleculer {
 	/**
@@ -1340,21 +1341,23 @@ declare namespace Moleculer {
 			sender: string | null;
 		}
 
+		type packetType = | PACKET_UNKNOWN
+			| PACKET_EVENT
+			| PACKET_DISCONNECT
+			| PACKET_DISCOVER
+			| PACKET_INFO
+			| PACKET_HEARTBEAT
+			| PACKET_REQUEST
+			| PACKET_PING
+			| PACKET_PONG
+			| PACKET_RESPONSE
+			| PACKET_GOSSIP_REQ
+			| PACKET_GOSSIP_RES
+			| PACKET_GOSSIP_HELLO;
+
 		interface Packet {
-			type:
-				| PACKET_UNKNOWN
-				| PACKET_EVENT
-				| PACKET_DISCONNECT
-				| PACKET_DISCOVER
-				| PACKET_INFO
-				| PACKET_HEARTBEAT
-				| PACKET_REQUEST
-				| PACKET_PING
-				| PACKET_PONG
-				| PACKET_RESPONSE
-				| PACKET_GOSSIP_REQ
-				| PACKET_GOSSIP_RES
-				| PACKET_GOSSIP_HELLO;
+			type: packetType
+
 			target?: string;
 			payload: PacketPayload;
 		}
@@ -1663,6 +1666,59 @@ declare namespace Moleculer {
 				plainError: PlainMoleculerError,
 				payload: GenericObject
 			): Error | undefined;
+		}
+	}
+
+	/**
+	 * you can use multiple modifier with a dot
+	 * e.g. `black.bgRed.bold`
+	 */
+	type KleurColor = keyof Kleur | string
+
+	namespace Middlewares {
+		namespace Debugging {
+			type actionLoggerOptions = {
+				logger?: LoggerInstance,
+				logLevel?: LogLevels,
+				logParams?: boolean,
+				logResponse?: boolean,
+				logMeta?: boolean,
+
+				folder?: string | null,
+				extension?: string,
+
+				colors?: {
+					request?: KleurColor,
+					response?: KleurColor,
+					error?: KleurColor
+				},
+				whitelist?: Array<string>
+			}
+			const ActionLogger = (options?: actionLoggerOptions) => Middleware
+			type transitLoggerOptions = {
+				logger?: LoggerInstance,
+				logLevel?: LogLevels,
+				logPacketData?: boolean,
+
+				folder?: string | null,
+				extension?: string,
+
+				colors?: {
+					receive?: KleurColor,
+					send?: KleurColor
+				},
+
+				packetFilter?: Array<Packets.packetType>
+			}
+			const TransitLogger = (options?: transitLoggerOptions) => Middleware
+		}
+		namespace Transmit {
+			type compressionOptions = {
+				method?: "deflate" | "deflateRaw" | "gzip",
+				threshold?: number | string
+			}
+			const Compression = (options?: compressionOptions) => Middleware
+			const Encryption = (password: string, algorithm: string, iv: string | Buffer) => Middleware
 		}
 	}
 


shawnmcknight (commented):
A couple things here:
1. What's the use case for making this change?
2. Although there are many places in the existing types where `namespace` is used, its a pattern I'd prefer to avoid in the future.  Can the changes you're making be accomplished without using them?
3. Types should always be PascalCase and not camelCase.  There are several changes here using camelCase.

thib3113:
@shawnmcknight 

> 1. What's the use case for making this change?

It allow to add middleware in moleculer.config.ts, without using ts-ignore
image


> 2. Although there are many places in the existing types where `namespace` is used, its a pattern I'd prefer to avoid in the future.  Can the changes you're making be accomplished without using them?

I will check

> 3. Types should always be PascalCase and not camelCase.  There are several changes here using camelCase.

Ok, I will check

thib3113:
when I want to use the middleware in my moleculer.config.ts now : 
image

thib3113:
@shawnmcknight any news on this ?

index.d.ts
```diff
@@ -1342,21 +1343,24 @@ declare namespace Moleculer {
 			sender: string | null;
 		}
 
+		type packetType =
```
shawnmcknight, line 1346: Types should be `PascalCase`

index.d.ts
```diff
@@ -1668,6 +1672,54 @@ declare namespace Moleculer {
...
+			receive?: KleurColor;
+			send?: KleurColor;
+		};
+		packetFilter?: Array<Packets.packetType>;
+	};
+	type CompressionOptions = {
+		method?: "deflate" | "deflateRaw" | "gzip";
+		threshold?: number | string;
+	};
+	export const Middlewares: {
```
shawnmcknight, line 1712: I apologize for sending you down the wrong path earlier.  I've subsequently learned that namespaces are appropriate for modeling CJS libraries.  `Debugging`, `Transmit`, and `Middlewares` should all be namespaces.

index.d.ts
```diff
@@ -1668,6 +1672,54 @@ declare namespace Moleculer {
 		}
 	}
 
+	/**
+	 * you can use multiple modifier with a dot
+	 * e.g. `black.bgRed.bold`
+	 */
+	type KleurColor = keyof Kleur | string;
+
+	type ActionLoggerOptions = {
```
shawnmcknight, line 1681: objects should be modeled as `interface` rather than `type` because it allows for augmentation by plugins and consumers.

index.d.ts
```diff
@@ -1668,6 +1672,54 @@ declare namespace Moleculer {
...
+		threshold?: number | string;
+	};
+	export const Middlewares: {
+		Debugging: {
+			ActionLogger(options?: ActionLoggerOptions): Middleware;
+			TransitLogger(options?: TransitLoggerOptions): Middleware;
+		};
+		Transmit: {
+			Compression(options?: CompressionOptions): Middleware;
+			Encryption(password: string, algorithm: string, iv: string | Buffer): Middleware;
```
shawnmcknight, line 1719: Looks like algorithm is used to call into `crypto.createCipheriv` and `crypto.createCipher` which accepts a type of `crypto.CipherCCMTypes`.

shawnmcknight: `iv` is of type `crypto.BinaryLike`.

index.d.ts
```diff
@@ -1668,6 +1672,54 @@ declare namespace Moleculer {
...
+		logResponse?: boolean;
+		logMeta?: boolean;
+		folder?: string | null;
+		extension?: string;
+		colors?: {
+			request?: KleurColor;
+			response?: KleurColor;
+			error?: KleurColor;
+		};
+		whitelist?: Array<string>;
```
shawnmcknight, line 1694: Please use `string[]` instead of `Array<string>`.

index.d.ts
```diff
@@ -1668,6 +1672,54 @@ declare namespace Moleculer {
...
+		logger?: LoggerInstance;
+		logLevel?: LogLevels;
+		logPacketData?: boolean;
+		folder?: string | null;
+		extension?: string;
+		colors?: {
+			receive?: KleurColor;
+			send?: KleurColor;
+		};
+		packetFilter?: Array<Packets.packetType>;
```
shawnmcknight, line 1706: Two things here:
1) Since you've moved `PacketType` into its own type, you can just reference that without the leading `Packet.`
2) Please use `PacketType[]` instead of `Array<...>`

index.d.ts
```diff
@@ -1668,6 +1672,54 @@ declare namespace Moleculer {
 		}
 	}
 
+	/**
+	 * you can use multiple modifier with a dot
+	 * e.g. `black.bgRed.bold`
+	 */
+	type KleurColor = keyof Kleur | string;
```
shawnmcknight, line 1679: Is union with `string` necessary here, or should the colors all be from `kleur`?

shawnmcknight (changes requested):
Overall this looks good.  I had a few requests for adjustment.

index.d.ts
thib3113: KleurColor can be linked together . like I said on the comment . 

So : `${Kleur.black}` is valid, `${Kleur.black}.${Kleur.bgRed}` too, etc

shawnmcknight: I wouldn't think it worth it to go down the path of making a recursive type.

Unfortunately, once you union the `keyof Kleur` with `string` its going to widen to `string`, effectively rendering the `keyof Kleur` moot.  If allowing `string` is important, you can likely just allow `string` without the `keyof Kleur` because it ends up in the same place anyway.

index.d.ts
thib3113: PacketType is in the Packet namespace . and not the packetFilter, So I need to use the `Packet.PacketType` notation

index.d.ts
thib3113: there is even more complexe, it can be `crypto.CipherCCMTypes` / `crypto.CipherOCBTypes` / `crypto.CipherGCMTypes`, also the password length will depend on the algorithm . 

And more, this middleware is a bad one . The encryption without IV is deprecated (not really secure), and reusing the same IV for all messages is also not secure . ( IV need to be included in the message, and change on every messages https://medium.com/swlh/an-introduction-to-the-advanced-encryption-standard-aes-d7b72cc8de97#:~:text=Initialization%20vector%20(IV),is%20an%20added%20security%20layer. ) .

index.d.ts
thib3113: the union, allow intellisense without limiting too : 
image

index.d.ts
thib3113: not a problem for me but : 

this options can't be augmented (because middleware can't be augmented ? or I didn't known ? ), this options are totally dependends of the middleware ? ( also, this didn't change lot of things, so, not a problem to migrate to interfaces )

thib3113:
Please @shawnmcknight / @icebob can you review this PR ? it is waiting for nearly a year, where PR https://github.com/moleculerjs/moleculer/pull/1246 with same changes was validated in 2 days (litterally the PR just re-do a part of this one) ...

PR #1164, opened 2022-12-11
(types): migrate types to interface

thib3113:
## :memo: Description

I got the following problem : 

I can't extends the params of an event
I want to create a custom property on events `debounce2`
```
events: {
    myEvent: {
        debounce2: 123,
        handler: async (ctx: Context<ICommandParams>) => true
    }
}
```
But, I can't do this with typescript, because the list of options is locked !
I can't extend it with
```typescript
declare module 'moleculer' {
    interface ServiceEvent {
        debounce2?: number;
    }
}
```

because ServiceEvent is declared as a type, and not an interface . So, after lot of search, I found this : [differences-between-type-aliases-and-interfaces](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) . And it say using type will block the modifications (and so a middleware can't add options) . 

And this is a problem with middlewares . 

So, I update all types that want to be extended in the future, just by replacing types by interfaces + I update some parts about the thisType (I add the this in the handler) + run linter (so some lines are only reformated).

### :gem: Type of change

### :scroll: Example code
this allow middlewares developper to do : 
```typescript
declare module 'moleculer' {
    interface ServiceEvent {
        debounce2?: number;
    }
}
``` 

and so : 
image

- [X] I run auto typescripts tests
- [X] I try to build my projects using moleculer with this

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -162,6 +162,8 @@ declare namespace Moleculer {
 		startSpan(name: string, opts?: GenericObject): Span;
 	}
 
+	// allow to extend TracingActionTags
+	interface TracingActionTagsExtension {}
 	type TracingActionTagsFuncType = (ctx: Context, response?: any) => GenericObject;
 	type TracingActionTags =
 		| TracingActionTagsFuncType
@@ -169,15 +171,17 @@ declare namespace Moleculer {
 				params?: boolean | string[];
 				meta?: boolean | string[];
 				response?: boolean | string[];
-		  };
+		  } & TracingActionTagsExtension;
 
+	// allow to extend TracingEventTags
+	interface TracingEventTagsExtension {}
 	type TracingEventTagsFuncType = (ctx: Context) => GenericObject;
 	type TracingEventTags =
 		| TracingEventTagsFuncType
 		| {
 				params?: boolean | string[];
 				meta?: boolean | string[];
-		  };
+		  } & TracingEventTagsExtension;
 
 	type TracingSpanNameOption = string | ((ctx: Context) => string)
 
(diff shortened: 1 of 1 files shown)


shawnmcknight (commented):
While I agree that many scenarios where `type` is currently used should be changed to `interface`, there are several changes here which I think will break existing behaviors.

For instance, some places have been changed from type unions to interfaces extending function types.  This will not have the desired outcome because the type will then require implementation of the function in all cases, rather than allowing either a function or an object.  

Additionally, there are lots of changes around the use of `this` which are very tricky to get right.  The use of `ThisType` in these locations is deliberate.

The better approach to these changes is to leave type unions and `ThisType` unchanged in as many places as possible.  Any place where there is an object defined, it would be good to switch to an `interface` and then reuse that interface name in those type aliases.  Then for augmentation, the interfaces can be extended augmented as necessary.

thib3113:
> For instance, some places have been changed from type unions to interfaces extending function types. This will not have the desired outcome because the type will then require implementation of the function in all cases, rather than allowing either a function or an object.

Exact, I just checked, and it's a problem . ( I just based my PR on tests, and on my project )
 
> Additionally, there are lots of changes around the use of `this` which are very tricky to get right. The use of `ThisType` in these locations is deliberate.
yes, but in fact, the "this", is not defined on the object point of view, only in functions ? ( so finally this need to be defined on functions, and not on the object ? )


> The better approach to these changes is to leave type unions and `ThisType` unchanged in as many places as possible. Any place where there is an object defined, it would be good to switch to an `interface` and then reuse that interface name in those type aliases. Then for augmentation, the interfaces can be extended augmented as necessary.
Yes, can be an option too .

thib3113:
@shawnmcknight So, I reset it, and create interfaces, ending with Extension .

This interfaces allow to extends types easy .
I add more *extension that I really need, but It's to allow more extensions in the future, without always recreating a PR . 

Also, I move some "middlewares" options in *extension, but it's only to show an examples for futures middlewares developpers .

thib3113:
@icebob yes, it's good . Thank you @shawnmcknight !

PR #1172, opened 2022-12-28
Fix merge schemas

0x0a0d:
if `mixinSchema` is null, `res[key]` will throw error
if `svcSchema` is null, `Object.keys(mods)` will throw error


icebob (approved):
Great, thanks!

PR #1179, opened 2023-01-22
Improve lifecycle handler types

shawnmcknight:
## :memo: Description

This PR improves the `Service` and `ServiceBroker` lifecycle handler type definitions in the following ways:

#### Service Lifecycle Handlers

1. The service lifecycle methods are called with a `this` value of the `Service`.  The type was updated to have a `this` value of the `Service`.
2. The `created` handler was not consistent in structure with the `started` and `stopped` handlers.  The `created` handler was defined inline while the `started` and `stopped` handlers were aliased.  This was adjusted so that both are now aliased.
3. The `StartedStoppedHandler` type was rather specific in name, so to make it more generic it was renamed to `ServiceAsyncLifecycleHandler`.  The counterpart for the `created` handler was then a `ServiceSyncLifecycleHandler`.  These can then be reused as necessary without any naming confusion.
4. A recent PR (#1173) had adjusted the previously named `StartedStoppedHandler` to allow returning `Promise<void[]>`.  However, that is not the intent of these lifecycle methods -- they should return `void | Promise<void>`.  Any other type being returned is not appropriate, including `Promise<void[]>`.  The intent was to allow a return of `Promise.all[...]` but the preferred approach would be something like:

```
async started() {
  await Promise.all([...]);
  // implicit return of void
}
```

#### Broker Lifecycle Handlers
The main issue with the broker lifecycle handlers is that the `started` and `stopped` handlers do not allow returning `Promise<void>` and are expecting a return of `void`.  Changes have been made to address that deficiency as well as make the types used consistent with the changes that were just made to the service lifecycle handlers.

#### Other Changes
Any other changes were unintentional and only a result of prettier auto-formatting making adjustments.

### :gem: Type of change

Created a test repo and set it up to work with these type changes.  Confirmed the following:
1. The service broker `started` and `stopped` handlers now allow passing a Promise returning function.
2. The service `created`, `started`, and `stopped` handlers now infer `this` as the `Service` type rather than the `ServiceSchema` type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -334,11 +334,7 @@ declare namespace Moleculer {
 		generateSnapshot(): HistogramMetricSnapshot[];
 
 		static generateLinearBuckets(start: number, width: number, count: number): number[];
-		static generateExponentialBuckets(
-			start: number,
-			factor: number,
-			count: number
-		): number[];
+		static generateExponentialBuckets(start: number, factor: number, count: number): number[];
 	}
 
 	namespace MetricTypes {
@@ -584,7 +580,7 @@ declare namespace Moleculer {
 		call<TResult, TParams>(
 			actionName: string,
 			params: TParams,
-			opts?: CallingOptions,
+			opts?: CallingOptions
 		): Promise<TResult>;
 
 		mcall<T>(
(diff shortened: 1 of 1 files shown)


icebob:
ping @danieledraganti

danieledraganti:
@icebob fine by me; I proposed that only because I found that form (`return Promise.all(...)`) in the documentation somewhere when I set the service up first. What's proposed by @shawnmcknight makes sense; it should return the state of the event.

icebob (approved):

shawnmcknight:
I made one additional tweak to ensure that the generic for the settings schema type was passed through to the lifecycle handlers.

PR #1181, opened 2023-01-30
feat(runner): load service by matched path

0x0a0d:
I have written about this problem with env `SERIVCES`: moleculerjs/site#160
I think SERVICES should first prefer to `service's name`
This pull will allow loading service from SERVICES if any service's file ends with `{SERVICE_ENV_VALUE}.service.[tj]s`

### DIFF
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -387,6 +387,7 @@ export default class MoleculerRunner {
 
 		if (patterns.length > 0) {
 			let serviceFiles = [];
+			const allServiceFiles = glob(path.join(svcDir, fileMask), { absolute: true });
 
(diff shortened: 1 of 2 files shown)


icebob:
Could you write an exact example where the original code doesn't load the desired files?

0x0a0d:
Example:
```bash
.
└── src/
    └── services/
        └── counter/
            ├── counter-peter.service.js
            └── counter-marry.service.js
```
With SERVICEDIR=src/services
- `SERVICES=counter-marry`, there are no service found
- But `SERVICES=counter` will load both `counter-peter`, `counter-marry` (all `*.service.js` in `counter` folder)

PR #1188, opened 2023-03-04
runner multi env load

0x0a0d:
Load specified .env files using glob patterns that are separated by commas. For example, '-E .env,.env.dev.*' loads the .env file and all .env.dev.* files.


0x0a0d:
@icebob this pull is compatible with current version, it both supports single .env file or multiple .env files, could you merge it?

PR #1192, opened 2023-03-08
Allow opt out of redlock in redis cacher

shawnmcknight:
## :memo: Description

Currently, the Redis cacher will always try to instantiate a redlock instance if the dependency is able to be resolved.  Currently, moleculer is only compatible with `^4.0.0`.  The current version of redlock is `5.0.0-beta.2` and moleculer is not compatible with it.  However, if moleculer users with to use the current version of redlock while using the Redis cacher, they will experience errors when the cacher gets instantiated that prevent the broker from starting.  This happens whether consumers wish to make use of the redlock functionality in moleculer or not.

This PR allows consumers to opt-out of using redlock in the Redis cacher by setting the `opts.redlock` to `false`.  This was done to avoid any breaking changes.

Additionally, this PR added a little more safety around use of the `lock` and `tryLock` methods in the cacher.  If redlock is not installed or was disabled, those methods would currently error because they would be accessing something not instantiated.  Those methods will now log an error and resolve instead.

### :gem: Type of change

New unit tests were added and existing unit tests continue to pass.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -815,7 +815,10 @@ declare namespace Moleculer {
 		 * @param mixinSchema Mixin schema
 		 * @param svcSchema Service schema
 		 */
-		mergeSchemas(mixinSchema: Partial<ServiceSchema>, svcSchema: Partial<ServiceSchema>): Partial<ServiceSchema>;
+		mergeSchemas(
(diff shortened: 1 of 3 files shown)


icebob (approved):
Great job, thanks!

PR #895, opened 2021-03-24
fix: datadog tracer scope

ngraef:
## :memo: Description

- Support dd-trace's AsyncLocalStorage and AsyncResource scope manager implementations (introduced in [v0.27.0](https://github.com/DataDog/dd-trace-js/releases/tag/v0.27.0) and [v0.28.0](https://github.com/DataDog/dd-trace-js/releases/tag/v0.28.0)) and clean up the AsyncHooks implementation.
- Fix activation of moleculer spans in dd-trace, allowing for propagation to external modules. This also allows for log injection to [connect logs and traces](https://docs.datadoghq.com/tracing/connect_logs_and_traces/).
- Allow moleculer to continue a trace that was started in an external module. For example, this will show the moleculer-web handler as a child span of the node http listener.

### :dart: Relevant issues

Resolves #690
Resolves  #889

### :gem: Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## :vertical_traffic_light: How Has This Been Tested?

- [x] Unit and integration tests added
- [x] Tested an environment with 5 distributed services (each running in separate containers) sending traces to Datadog. The sample screenshots below show cross-service calls, custom spans in an action handler, and modules auto-instrumented by dd-trace (http-client, tcp, dns, fs, redis, postgres).

###### A simple call through moleculer-web to a remote action that makes some HTTP calls and interacts with a postgres database
Screen Shot 2021-04-18 at 11 39 09 AM

###### A call that includes some nested remote action calls, one of which calls redis and returns an error
Screen Shot 2021-04-18 at 11 40 06 AM

###### The same call as the previous example, but with a remote service subscribing to an event it emits
Screen Shot 2021-04-18 at 11 16 12 PM

###### The initial call to a lazy-loaded module with a bunch of fs reads needed to require the library
Screen Shot 2021-04-18 at 11 40 33 AM

###### An action handler with some custom spans via `ctx.startSpan`
Screen Shot 2021-04-18 at 11 41 49 AM

## :checkered_flag: Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] **I have added tests that prove my fix is effective or that my feature works**
- [x] **New and existing unit tests pass locally with my changes**
- [x] I have commented my code, particularly in hard-to-understand areas

PR #1206, opened 2023-05-05
fix: utilize dd-trace's provided scope management

dowster:
## :memo: Description

dd-trace@1.5.1 was not receiving the scope properly with the code in place. Any auto-instrumented traces dd-trace generated would have unique trace IDs and no parent links. This change will fix the scope handling. Like the existing implementation, this will only work with [dd-trace@1.5.1 `async_hooks` scope management](https://github.com/DataDog/dd-trace-js/blob/v1.5.1/packages/dd-trace/src/scope/async_hooks.js).

@ngraef had made a similar patch with PR #895. He also included support for other scope managers. I'd rather not implement that now as we're looking at working with OpenTelemetry libraries for further tracing enhancements and possibly switching from dd-trace.

### :gem: Type of change

### :scroll: Example code
```js

``` 

This change was deployed to our testing environment for a few different services. I verified that trace propagation was working correctly with dd-trace's Postgres, Redis, HTTP, TCP, and file system instrumentation. 

We are using the dd-trace@1.5.1 library.


src/tracing/exporters/datadog.js
```diff
@@ -157,17 +157,10 @@ class DatadogTraceExporter extends BaseTraceExporter {
...
 
-		// Activate span in Datadog tracer
-		const asyncId = asyncHooks.executionAsyncId();
-		this.ddScope._spans = this.ddScope._spans || {};
-		const oldSpan = this.ddScope._spans[asyncId];
-
-		this.ddScope._spans[asyncId] = ddSpan;
+		this.ddTracer.scope()._enter(ddSpan);
 
 		span.meta.datadog = {
```
OutdatedVersion, line 162: Any idea on if this meta property is considered internal? I could see the `asyncId` being useful in some middlewares

icebob:
How this changes work with older versions? Because the Moleculer 0.14 supports the following dd-trace versions: `"dd-trace": "^0.33.0 || ^0.34.0 || ^0.35.0 || ^0.36.0 || >=1.0.0 <1.6.0",`

If it doesn't work with older and we should unsupport them, it means, this PR can be released only in the next major 0.15 version of Moleculer

dowster:
> How this changes work with older versions? Because the Moleculer 0.14 supports the following dd-trace versions: `"dd-trace": "^0.33.0 || ^0.34.0 || ^0.35.0 || ^0.36.0 || >=1.0.0 <1.6.0",`
> 
> If it doesn't work with older and we should unsupport them, it means, this PR can be released only in the next major 0.15 version of Moleculer

I can add a check my change to utilize existing logic if dd-trace is less than 1.5.1, and utilize my change if it is 1.5.1 since I have tested on 1.5.1. 

On the other hand, could we push this change out with version 0.15.0 and remove support for older dd-trace versions? I think allowing more breaking changes / frequent version bumps could possibly help move the project along more quickly.

icebob:
Yeah, this is the plan. For this purpose, could you change this PR with removing old dd-trace versions and change the target branch from master to `next`, please?

dowster:
I rebased the PR onto the `next` branch, updated the peerDependency and devDependency of datadog to the latest major, and updated a test which broke due to a change inside dd-trace. 

I figured going to dd-trace@4 and dropping all others should be acceptable as older versions of dd-trace support node@12 and node@14 while both moleculer/next and dd-trace@4 require node>=16. 

I still need to do some integration testing with this build in my cluster to confirm nothing breaks. This version of dd-trace includes [moleculer integration](https://github.com/DataDog/dd-trace-js/tree/master/packages/datadog-plugin-moleculer), so I want to be sure we don't start seeing double counted spans. In a future change, we may be able to remove most of the exporter logic in the moleculer trace exporter and rely completely on dd-trace's instrumentation.

icebob:
>  In a future change, we may be able to remove most of the exporter logic in the moleculer trace exporter and rely completely on dd-trace's instrumentation.

In the future I will remove the built-in tracing solution and using only OpenTelemetry: https://github.com/moleculerjs/moleculer/discussions/1125

icebob:
@dowster any news about it? I'm working on to update all dependencies and dd-trace is the latest one but it conflicts with your PR. I've checked your PR and I've got error messages. Could you work on it to fix it and I can merge it to the next branch?

dowster:
> @dowster any news about it? I'm working on to update all dependencies and dd-trace is the latest one but it conflicts with your PR. I've checked your PR and I've got error messages. Could you work on it to fix it and I can merge it to the next branch?

I might not be able to get back to this for another week. I'll also need to set up some new test scenarios, since I won't have access to my current test environments or datadog accounts. I'll post back here later to update if it's still feasible for me to complete this.

icebob:
Plz post back if you are not feasible because in this case, I should work on it.

PR #1208, opened 2023-05-09
Improve logger types

shawnmcknight:
## :memo: Description

The current types for the `LoggerFactory` and `Base` Logger are incorrect in some cases or not specific enough in others.  Multiple methods are accepting `GenericObject` instead of the more accurate `LoggerBindings` type and have been updated accordingly.  Additionally, rather than allowing `string` as a type for the log levels return value, the more specific `LogLevels` type was used.  Finally, the return types for `getLogHandler` was updated to reflect the correct return type of the function.

### :gem: Type of change

I've patched my local moleculer installation with these changes and implemented a custom logger using those fixed types.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,8 +21,8 @@ declare namespace Moleculer {
 		constructor(broker: ServiceBroker);
 		init(opts: LoggerConfig | LoggerConfig[]): void;
 		stop(): void;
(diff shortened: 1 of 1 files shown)


icebob (approved):
Cool, thanks!

Issue #747, opened 2020-05-22
[lodash] Remove usage of _.compact


abdavid:
See #433 for context.

**Describe the solution you'd like**
Remove usage of `_.compact` throughout Moleculer. 

**Describe alternatives you've considered**
Thesis is that it would be sufficient to use `filter(Boolean)`
E.g.

```
const filteredList = [true, false, 0, NaN, undefined, "", null].filter(Boolean)
```

**Additional context**
https://github.com/moleculerjs/moleculer/search?q=_.compact&unscoped_q=_.compact

abdavid:
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_compact

PR #1209, opened 2023-05-12
Remove usage of _.compact

MrAndrewKirillov:
### :dart: Relevant issues

[issue:747](https://github.com/moleculerjs/moleculer/issues/747)

### :gem: Type of change

### DIFF
--- a/src/logger-factory.js
+++ b/src/logger-factory.js
@@ -54,7 +54,7 @@ class LoggerFactory {
 				opts = [opts];
 			}
 
-			this.appenders = _.compact(opts).map(o => {
+			this.appenders = opts.filter(Boolean).map(o => {
 				// Built-in shorthand
 				if (isString(o))
 					return Loggers.resolve({ type: o, options: { level: globalLogLevel } });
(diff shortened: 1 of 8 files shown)


icebob (changes requested):
Would be better if you add a `compact` method to [utils ](https://github.com/moleculerjs/moleculer/blob/master/src/utils.js)(similar to other lodash compatible functions) and use it in the codes.

MrDroid1:
@icebob I think creating method for `compact` is overhead, cause function too small and we haven't to use lodash style. May you describe your point of it, it's easy to add method, but I realy don't understend point of it.

icebob:
The lodash `compact` as the input is not array, or null or undefined. So when I'm using I don't care to check the parameter before add it to function because it will handle and always returns an array. The current solution doesn't handle it, but in a utils function we can do this check as well.

image

MrDroid1:
> The lodash compact as the input is not array, or null or undefined.

Yeah, but this behavior not expected and lodash documentation says you have to use array as a parameter to remove falsy values. I think we should not rely on the behavior of the method, but should proceed from the needs, that is, the method should remove falsy values from the array. `fillter(Boolean)` do that, remove falsy values from array. 

> So when I'm using I don't care to check the parameter before add it to function because it will handle and always returns an array

In my opinion, it's a bad idea to rely on a single method and the best solution would be to use monosemantic values and methods using monosemantic values and responses.

PR #1218, opened 2023-06-12
Update hot-reload.js, Avoiding the occurrence of "Maximum call stack size exceeded " on  hot-reload

zhangfisher:
## :memo: Description

Error occurred when enabling -- hot

 "Maximum call stack size exceeded "

I found out I was checking the node_ The occurrence of modules dependency


icebob:
Could you make a reproduce repo for the issue? Because we can't filter out node_modules because the installed moleculer modules and the core module as ServiceBroker (which loads the services) are placed into the node_modules as well.

zhangfisher:
This problem is difficult to build repo, I guess it's due to the use of pnpm

PR #1221, opened 2023-06-16
Fix: prometheus reporter to accept host option

MichaelErmer:
## :memo: Description

add host option, because listen with only port will lead to node only listening on ipv6 address if ipv4 and ipv6 are enabled.

https://stackoverflow.com/questions/47797322/node-js-server-only-listening-on-ipv6

### :gem: Type of change


MichaelErmer:
> Please fix the CI.

How? it seems to be some infrastructure problem:
```
  ● Test suite failed to run

    getaddrinfo EAI_AGAIN test-host
```

MichaelErmer:
> I don't think so, because the master branch passed and if I fetch this PR, the error appears on my local dev as well.

Yep, funny issue with node not accepting a host when using the port: 0 feature, fixed!

icebob:
Nice catch, thanks!

PR #1223, opened 2023-07-02
Feature: Add the ability to ignore services

mohsen-karami:
There are situations in that the user may want to ignore some services; this feature allows them to do this by simply adding them as an array and putting it as the third parameter of the loadServices method.

## :memo: Description

In our example, one service needed to install the TensorFlow package, which was incompatible with the hardware of some developers. Since they don’t code on the service and it was unnecessary for running the app, I’ve added this feature, thus they can simply ignore the service. In our case, since we still want the service to be available for others, I defined an environment variable in the `.env` file and then wrote an if statement to ignore the intended services only if their relevant environment variable (in our case, `EXCLUDED_SERVICES`) exists.

### :gem: Type of change


AndreMaz:
Hi @mohsen-karami 

You can ignore specific services via env vars.  like:
```bash
SERVICES=src/services,!**/api.service.js // will load everything in src/services expect for the "api.service"
```

This approach is not enough for your use case?

mohsen-karami:
Hey @AndreMaz

Yeah, thanx.
Your approach works great if we run the project using the moleculer-runner, but since we’re running multiple instances of Moleculer and want to have even more control over them, we had to create the ServiceBroker manually.

I think this implementation is convenient for any other who runs the broker manually (for any possible reason!).

icebob:
Thank you for your job, but as André said, this functionality is covered in Runner with SERVICE env var, and if we implement the same logic into the broker as well, in the future it will confuse us and make it hard to maintain all implementation. 

So at this point, I won't merge it because you can do it inside your project, if you override the `loadServices` method or add a new custom method in ServiceBroker with you logic.

By the way, if we faced that many users need this feature, we will consider it.

mohsen-karami:
OK, thanx @icebob for the review.

you’re right, if it’s only us needing this feature, considering that it may add complexity and confusion to the project, making maintenance more difficult for you, then I guess better to follow your approach.

icebob:
@mohsen-karami Thank you for your understanding

PR #1229, opened 2023-07-19
Bugfix: started not being called when createService is used

MichaelErmer:
* fix: service initialization/started not being called
 
This change prevents the bug that started is not called for services created by broker.createService while the initial services (loaded from files) are still being "started" (some might take longer with some external connections being setup etc.) OR even the broker.createService being called from inside a service's started handler.

### :gem: Type of change

### :scroll: Example code
```js
const dbService: ServiceSchema = {
  name: "test",
  started: async() => {
    this.broker.createService({
      name: "never-being-started",
      started: () => console.log("not gonna happen")
    });
    await new Promise(resolve => setTimeout(resolve, 1000));
  }
};
export = dbService;
``` 

- [x] Local Testing


MichaelErmer:
Workaround until this is merged:
```
middlewares: [
        IdempodencyMiddleware.middleware(),
        {
            starting: (broker: ServiceBroker) => {
                broker.started = true;
            },
        },
    ],
```

src/service-broker.js
```diff
@@ -186,6 +186,9 @@ class ServiceBroker {
 			// Broker started flag
 			this.started = false;
 
+			/** @type {Boolean} Broker is starting inital services flag*/
+			this.servicesStarted = false;
```
icebob, line 190: The variable name is not represent the exact state because it means services is **starting** right now, so would be better the `servicesStarting` name. But after that, it should set to false when all services started.

src/service-broker.js
```diff
@@ -834,8 +839,8 @@ class ServiceBroker {
 			service = new this.ServiceFactory(this, schema, schemaMods);
 		}
 
-		// If broker has started yet, call the started lifecycle event of service
-		if (this.started) this._restartService(service);
+		// If broker has began to start its initial services yet, call the started lifecycle event of service
+		if (this.servicesStarted) this._restartService(service);
```
icebob, line 843: the `if (this.started || this.servicesStarting)` would be better.

src/service.js
```diff
@@ -281,38 +281,44 @@ class Service {
...
+				.then(() => {
+					// Register service
+					return this.broker.registerLocalService(this._serviceSpecification);
+				})
+				.then(() => {
+					return this.broker.callMiddlewareHook("serviceStarted", [this]);
+				})
+				.then(() => this.logger.info(`Service '${this.fullName}' started.`));
+		}
+		return this._starting;
```
icebob, line 321: This variable is not nulled after the start process finished.

icebob (changes requested):
Your use-case is not typical and not a recommended way to start services. So, I'm not sure this PR is necessary for the framework. 

@AndreMaz what do you think?

P.S.: It's missing unit tests and while it modified the service loading, it should be covered by integration tests, as well.

src/service.js
MichaelErmer: That should be no problem, the resolved promise will always return the same result, also if it would be nulled there's a chance `started` can be called again

MichaelErmer:
> Your use-case is not typical and not a recommended way to start services. So, I'm not sure this PR is necessary for the framework.
> 
> @AndreMaz what do you think?
> 
> P.S.: It's missing unit tests and while it modified the service loading, it should be covered by integration tests, as well.

We are using this to dynamically instantiate sequelize services based on a definition file, I don't think anything can break from these changes, but if you have an idea which test cases should be added, let me know, and I can implement them.

src/service-broker.js
MichaelErmer: updated

src/service-broker.js
MichaelErmer: updated.

src/service.js
MichaelErmer: i removed all changes to service._start / started, the broker changes are sufficient to cover the issue

MichaelErmer:
@icebob I removed all changes to the service class and added a test case that fails with the old implementation, so it tests the actual improvement.

src/service-broker.js
icebob: It's still missing.

icebob (approved):

PR #1243, opened 2023-09-10
fix memory leak

ghostiee:
Resume stream after all chunks are published to prevent memory leak

## :memo: Description

Memory leak scenario: we have two machines with each deployed a node. We use molecularr-web as a proxy to implement the file upload function. The files are first sent to the gateway and then forwarded to the service on the other machine using streams.
However, due to the file storage speed being slower than the internal network speed between the two machines, data starts to accumulate in the gateway.
As we had not handled this situation correctly before, it caused significant memory leaks in our production environment when users uploaded files. I identified the issue and fixed it.

### :dart: Relevant issues

None.

### :gem: Type of change

As the description above

### DIFF
--- a/src/transit.js
+++ b/src/transit.js
@@ -895,7 +895,7 @@ class Transit {
 						} else {
 							chunks.push(chunk);
 						}
-						for (const ch of chunks) {
+						const publishMap = chunks.map(ch => {
 							const copy = Object.assign({}, payload);
 							copy.seq = ++payload.seq;
 							copy.stream = true;
@@ -905,12 +905,13 @@ class Transit {
 								`=> Send stream chunk ${requestID}to ${nodeName} node. Seq: ${copy.seq}`
 							);
 
-							this.publish(new Packet(P.PACKET_REQUEST, ctx.nodeID, copy)).catch(
-								publishCatch
-							);
-						}
-						stream.resume();
-						return;
+							return this.publish(new Packet(P.PACKET_REQUEST, ctx.nodeID, copy));
+						});
+
+						return this.Promise.all(publishMap).then(
+							() => stream.resume(),
+							publishCatch
+						);
 					});
 
 					stream.on("end", () => {
@@ -1126,7 +1127,7 @@ class Transit {
 				} else {
 					chunks.push(chunk);
 				}
-				for (const ch of chunks) {
+				const publishMap = chunks.map(ch => {
 					const copy = Object.assign({}, payload);
 					copy.seq = ++payload.seq;
 					copy.stream = true;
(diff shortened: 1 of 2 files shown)


src/transit.js
```diff
@@ -905,12 +905,13 @@ class Transit {
 								`=> Send stream chunk ${requestID}to ${nodeName} node. Seq: ${copy.seq}`
 							);
 
-							this.publish(new Packet(P.PACKET_REQUEST, ctx.nodeID, copy)).catch(
```
icebob, line 908: I think, it would be much easier to add `await` to this row instead of changing the `for` iteration to a map and Promise.all

icebob (changes requested):

src/transit.js
ghostiee: Totally agree. I used to think that you want to send packets concurrently. if don't want to do this, it would be easier to add await in the for iteration

ghostiee: Oh, I remembered, this asynchronous iteration is in the event callback function. I'm afraid this is why you didn't add await before.

However, we can use the reduce method to make the publish process execute sequentially.

icebob: The original code sent all chunks in parallel, but I thought that your original problem was the parallelism or not?

ghostiee: No, The original issue was that the back pressure mechanism of the stream was not being correctly utilized, which caused memory leaks under certain conditions. 
In the original code, the stream was resumed before waiting for all the chunks to be fully published. As a result, the consumption progress of the stream did not match the publishing progress of the chunks. This caused the program to buffer excessive data in memory in some cases.

src/transit.js
```diff
@@ -1126,18 +1127,23 @@ class Transit {
...
 					chunks.push(chunk);
 				}
-				for (const ch of chunks) {
-					const copy = Object.assign({}, payload);
-					copy.seq = ++payload.seq;
-					copy.stream = true;
-					copy.data = ch;
 
-					this.logger.debug(`=> Send stream chunk to ${nodeID} node. Seq: ${copy.seq}`);
+				return this.Promise.all(
```
intech, line 1131: @icebob It's not a good solution to not control the size of chunks, as this can lead to excessive CPU usage in the `Promise.all` call.

icebob: what's the size which can cause problem?

intech: Based on my analysis, it seems most reasonable to derive the meaning from [`os.availableParallelism()`](https://nodejs.org/api/os.html#osavailableparallelism).

We are facing two issues: 
1. an unknown logic in `transitPublish` may cause synchronous blocking of operations in `Promise.all`.
2. We are unaware of the number of available system resources during the planning of `Promise.all` operations. We cannot control the peak load even if we had this information.

A small size of chunks for available resources is a bad solution, but for busy resources, it is a good solution.

PR #1246, opened 2023-09-22
Update index.d.ts with Middlewares.Transmit type definitions

0x0a0d:
Type definitions for the `Middlewares.Transmit` object


icebob (changes requested):
Please resolve the conflict.

icebob (approved):
Thanks!

PR #1248, opened 2023-09-25
imp(index.d.ts): middleware serviceCreating params

0x0a0d:
based on this
https://github.com/moleculerjs/moleculer/blob/master/src/middleware.js#L242


0x0a0d:
@icebob check failed because it could not start docker

icebob (approved):

PR #1258, opened 2023-11-17
Transit improve

0x0a0d:
while transporter instance is at 'connecting' state, if broker.stop is called, the transit will not ask transporter to disconnect.


icebob (approved):
Great, thanks

Issue #1241, opened 2023-09-02
Broker middleware `call` wrapper removes context data. Cannot use ctx.meta.


ggondim:
## Prerequisites

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository

## Current Behavior

When using a custom middleware in broker, specially when defing a `call` wrapper [exactly the way the official documentation indicates](https://moleculer.services/docs/0.14/middlewares#call-next), **context data appears to be removed.**

For example, moleculer-web's `context.meta.$statusCode` doesn't work when defining the middleware.

If you remove the middleware, everything works fine.

## Expected Behavior

Context data should be there after calling the action.

**moleculer-web's `context.meta.$statusCode` should work even with middlewares.**

### Steps to Reproduce

1. Create a `call` wrapper as a custom middleware [exactly the way the official documentation indicates](https://moleculer.services/docs/0.14/middlewares#call-next)

![image](https://github.com/moleculerjs/moleculer/assets/2074685/e9cf1895-4e0b-4224-9017-e51e4f070144)

2. Use `ctx.meta` between actions, such as setting `ctx.meta.$statusCode` using moleculer-web

![image](https://github.com/moleculerjs/moleculer/assets/2074685/0ef2b665-c605-476b-9d96-f42e32817ee3)

3. Call the action, the context will be lost

Take a look at the repro steps below.

### Reproduce code

Code is reproduced under [this repro repo](https://github.com/ggondim/moleculer-repro-repo).

1. Git clone it
2. `npm install && npm run dev`
3. `curl --location 'localhost:4987/api/statusCodeTest/testNotFound' -I`

You will see an HTTP 200 instead an HTTP 404 as defined in `statusCode.service.ts` action.

![image](https://github.com/moleculerjs/moleculer/assets/2074685/3fa1f543-d7dd-4cb2-8421-f32879ff7911)

### Context

* Moleculer version: `0.14.31`
* NodeJS version: `19.7.0`
* Operating System: `Ubuntu 22.04`

0x0a0d:
@icebob this bug happen because these
https://github.com/moleculerjs/moleculer/blob/master/src/middleware.js#L89
https://github.com/moleculerjs/moleculer/blob/master/src/middleware.js#L111

0x0a0d:
@icebob this bug happen because these
[...]

PR #1270, opened 2024-01-15
Add hook middlewares interceptors to preserve call context with call middlewares.

DonVietnam:
## :memo: Description

When using middleware for the call method, the call context was lost, as described in #1241  . This issue occurred because the context was passed through the mutation of the promise. Consequently, if the middleware awaited the result from the promise and returned a new promise, the execution context was lost.

In order to maintain backward compatibility, I propose the following solution:
We can wrap the middleware for specific hooks, such as call, in an interceptor function. The interceptor is responsible for setting the context value for the promise that will be returned from the user's middleware.

In this pull request, I have added interceptor functionality to the MiddlewareHandler class, as well as an interceptor method for the call hook in the ServiceBroker class.

### :dart: Relevant issues
#1241 

### :gem: Type of change

- [x] Integration middlware test "Test hook middleware interceptors".
- [x] Unit service broker test "Test broker.interceptCallMiddleware".

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -18,6 +18,8 @@ class MiddlewareHandler {
 		this.list = [];
 
 		this.registeredHooks = {};
+
+		this.middlewareInterceptors = {};
 	}
 
 	add(mw) {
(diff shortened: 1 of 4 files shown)


icebob:
Nice solution, thanks!

Issue #70, opened 2017-08-11
typescript support


alemp:
Hello!

Do you have any plan to support typescript?
I didn't see moleculer at @types.

Thanks in advance

icebob:
I don't have good knowledge in Typescript. I created a d.ts but I don't know it is correct. Could you help me about it?

PR #1272, opened 2024-02-05
feat: Expand the type for ServiceSchema to allow for typed lifecycle handlers

marceliwac:
This change expands the typings for the ServiceSchema, allowing it to accept the "service-this" type used by the lifecycle handlers in the typescript template. Please see [this issue](https://github.com/moleculerjs/moleculer-template-project-typescript/issues/70) for a more complete explanation.

## :memo: Description

Change the type of the ServiceSchema to accept two generic types, latter of which is new and used to provide types for the lifecycle methods (created, started, stopped).

The typescript template will require a minor change if this change is accepted.

### :dart: Relevant issues
[moleculer-template-project-typescript #70](https://github.com/moleculerjs/moleculer-template-project-typescript/issues/70)

### :gem: Type of change

The example code to verify that this change works is outlined in [moleculer-template-project-typescript #70](https://github.com/moleculerjs/moleculer-template-project-typescript/issues/70).

### DIFF
--- a/test/typescript/hello-world/greeter.service.ts
+++ b/test/typescript/hello-world/greeter.service.ts
@@ -4,7 +4,7 @@ import { Service } from "../../../";
 import { Context } from "../../../";
 import { ServiceSettingSchema, ServiceSchema } from "../../../";
 
-export interface ActionHelloParams {
+export interface ActionWelcomeParams {
 	name: string;
 }
 
@@ -24,6 +24,38 @@ type GreeterThis = Service<GreeterSettings> & GreeterMethods & GreeterLocalVars;
 
 const GreeterService: ServiceSchema<GreeterSettings, GreeterThis> = {
 	name: "greeter",
+
+	/**
+	 *  Hooks
+	 */
+	hooks: {
+		before: {
+			welcome(ctx: Context<ActionWelcomeParams>): void {
+				// console.log(`Service hook "before".`);
+				ctx.params.name = ctx.params.name.toUpperCase();
+			}
+		},
+		after: {
+			hello: "anotherHookAfter",
+			welcome: [
+				function (ctx: Context<ActionWelcomeParams>, res: any): any {
+					// console.log(`Service sync hook "after".`);
+					return res;
+				},
+				async (ctx: Context<ActionWelcomeParams>, res: any): Promise<any> => {
+					// console.log(`Service async hook "after".`);
+					return await Promise.resolve(res);
+				},
+				"anotherHookAfter"
+			],
+		},
+		error: {
+			welcome(ctx: Context<ActionWelcomeParams>, err: Error): void {
+				// console.log(`Service hook "error".`);
+				throw err;
+			}
+		}
+	},
 
 	/**
 	 * Settings
@@ -46,9 +78,22 @@ const GreeterService: ServiceSchema<GreeterSettings, GreeterThis> = {
 				method: "GET",
 				path: "/hello",
 			},
-			handler(this: GreeterThis/* , ctx: Context */): string {
+			handler(this: GreeterThis, ctx: Context): string {
 				return `Hello ${this.settings.defaultName}`;
 			},
+			hooks: {
+				after: [
+					(ctx: Context, res: any): any => {
+						// console.log(`Action sync hook "after".`);
+						return res;
+					},
+					async (ctx: Context, res: any): Promise<any> => {
+						// console.log(`Action async hook "after".`);
+						return await Promise.resolve(res);
+					},
+					"anotherHookAfter"
+				]
+			}
 		},
 
 		welcome: {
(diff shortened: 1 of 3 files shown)


test/typescript/hello-world/greeter.service.ts
```diff
@@ -2,88 +2,96 @@
 
 import { Service } from "../../../";
 import { Context } from "../../../";
-import { ActionHooks, ServiceHooks, ServiceHooksAfter, ServiceSchema } from "../../../";
```
icebob, line 5: Why you changed the test service file schema?

test/typescript/hello-world/index.ts
```diff
@@ -29,10 +29,8 @@ broker.loadService(path.join(__dirname, "greeter.service.ts"));
 
 		await broker.call("greeter.hello");
 		const res = await broker.call("greeter.welcome", { name: "Typescript" });
-		broker.logger.info("");
-		broker.logger.info("Result: ", res);
```
icebob, line 33: These changes will break the tests.

icebob (changes requested):

marceliwac:
@icebob , Apologies for any confusion!

I updated the test/typescript/hello-world/greeter.service.ts (i.e. service schema) to [the one used in the typescript template repository](https://github.com/moleculerjs/moleculer-template-project-typescript/blob/master/template/services/greeter.service.ts), since I couldn't find any in-code uses of the hooks or any tests related to that schema. I must be missing something, most likely the actual test that use this code. Could you please point me to the tests you're referring to?

I tried running all the tests specified in package.json which all pass fine (with the exception of memory leak tests, which I am guessing are unrelated).

icebob:
I've checked again the code and it looks the test can be fine, but you removed a lot of things from the original greeter service, e.g. the hooks which uppercaseing the context params. it would be good if you keep the same logic in the service file.

marceliwac:
@icebob You're absolutely right, sorry for that!

I reverted the deletions and tweaked them to comply with the service definition. I also added the implementation for the `uppercase()` method (which was declared as existing on the service but didn't) and replaced the direct call to `toUpperCase()` to a call to that method. Finally, I added the missing type delcaration for `this` where appropriate.

test/typescript/hello-world/greeter.service.ts
marceliwac: I updated the schema to add the missing hooks now.

icebob:
Great, thanks!

marceliwac:
@shawnmcknight

Apologies for the late reply and thanks for bringing this up!

You're right, these should in fact use `Service<S>` rather than `void`.

Ideally, if lifecycle handlers were nested within a separate property there would be a much cleaner solution that uses `ThisType<T>` utility like so: 

```ts
type ServiceSyncLifecycleHandler = () => void;
type ServiceAsyncLifecycleHandler = () => void | Promise<void>;

type ServiceLifecycle<T> = {
	created?: (ServiceSyncLifecycleHandler | ServiceSyncLifecycleHandler[]);
	started?: ServiceAsyncLifecycleHandler | ServiceAsyncLifecycleHandler[];
	stopped?: ServiceAsyncLifecycleHandler | ServiceAsyncLifecycleHandler[];
} & ThisType<T>;

interface ServiceSchema<S = ServiceSettingSchema, T = Service<S>> {
	name: string;
	version?: string | number;
	settings?: S;
	dependencies?: string | ServiceDependency | (string | ServiceDependency)[];
	metadata?: any;
	mixins?: Partial<ServiceSchema>[];
	methods?: ServiceMethods;
	hooks?: ServiceHooks;
        lifecycle?: ServiceLifecycle<T>;
	events?: ServiceEvents;
	[name: string]: any;
}
```

But because the lifecycle handlers are direct properties of the service, this makes things slightly messier. `ServiceSchema` is an interface and extending it with a type of `LifecycleHandlers<T>` (which is legal in TS >= 2.2) doesn't seem to apply the `ThisType<T>` correctly.

I also think that the generic `T` should be passed to the methods, events and handlers so we don't have to explicitly set `this` in each of the handlers. That matters less, because it still needs to be done if the function is defined outside of schema definition.

@icebob Any thoughts on this? Should I submit a PR?

icebob:
I have no idea due to lack of TS experience. @shawnmcknight ?

MichaelErmer:
what @marceliwac suggested would resolve our issues with the current version

marceliwac:
@shawnmcknight, @icebob, @MichaelErmer  I've pushed the [PR](#1299) to address this. I would appreciate a second pair of eyes sanity checking it if you could find some time. Hopefully, the added TS tests should prevent similar issues in the future.

I'm not 100% sure how to verify that `this` type of the lifecycle handlers (`created`, `started` and `stopped`) actually poitns to the instance of `Service` type, hence the use surrogate types (`ServiceSyncLifecycleHandler` and `ServiceAsyncLifecycleHandler`) in the tests. If you have better ideas, I'm open to suggestions.

PR #1274, opened 2024-02-14
Remove unnecessary clone in node update method

Ubitso:
## :memo: Description
We have a lot of instances of the app, with a bunch of services.
When app starts it took too long(4 sec on average) of active node cpu workload.
I've used the node debug to find a hot spot. 
That was a deepClone in update method of node.js.
It might be not necessary to clone the object because it comes from transport and it isn't used after.
I think current tests it has to be enough to prove that everything ok with the registry

### :gem: Type of change

It has to be enough current tests of node-catalog

### DIFF
--- a/src/registry/node.js
+++ b/src/registry/node.js
@@ -62,8 +62,7 @@ class Node {
 		this.client = payload.client || {};
 		this.config = payload.config || {};
 
-		// Process services & events (should make a clone because it will manipulate the objects (add handlers))
-		this.services = _.cloneDeep(payload.services);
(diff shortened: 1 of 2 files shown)


Ubitso:
image

Ubitso:
I've found cloneDeep was used to omit handler properties from services object. It breaks CBOR serializer e2e tests.

Ubitso:
@icebob Can you look this PR, please?

icebob (approved):
Good job, thanks

PR #1284, opened 2024-05-18
Fix runner use of moleculer.config function with default export

shawnmcknight:
## :memo: Description
When using moleculer-runner and a specified `moleculer.config.<ext>` file it should support both of the following:
- Use of default export syntax
- Exporting a function that returns a configuration object

The current flow of the runner is checking if the required `moleculer.config.<ext>` is a function prior to the logic which determines if the configuration file has a default export. This prevents the function from being executed if it is a default export.

This PR makes the following changes:
- Always return a configuration object from `loadConfigFile` even if a configuration file was not found
- Pass the configuration object returned from `loadConfigFile` to `mergeOptions`
- Use the passed configuration object in `mergeOptions` instead of `this.configFile`
- Eliminate use of `this.configFile`
- Change the order of processing in `loadConfigFile` so that the default export check precedes the function check and execution
- Adds support for `.mts` suffixes to the esm runner
- Adds new `moleculer.config.async.ts` and `moleculer.config.ts` files to the runner examples which validate typescript configurations and default exports for use with the runner

### :gem: Type of change

Tested using the newly created runner examples as well as the existing examples for both cjs and esm variants.

### DIFF
--- a/examples/runner/moleculer.config.async.ts
+++ b/examples/runner/moleculer.config.async.ts
@@ -1,0 +1,11 @@
+import fetch from "node-fetch";
+
+/**
+ * Test:
+ *
+ * 	npx ts-node -T bin\moleculer-runner.js -c examples\runner\moleculer.config.async.ts -r examples/user.service.js
+ */
+export default async function () {
+	const res = await fetch("https://pastebin.com/raw/SLZRqfHX");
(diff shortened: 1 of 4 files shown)


icebob (approved):
Good job, thanks!

PR #1286, opened 2024-06-06
Improve broker error handler types

shawnmcknight:
## :memo: Description
The `errorHandler` option that can be supplied to the service broker options has loose type definitions for the `info` parameter. In one case it is defined as `any`, and in the other it is defined as `GenericObject` (i.e. `Record<string,any>`). The signature of the callers to the `errorHandler` is variable, but it falls into two distinct categories: when called from the error handler middleware and when called from other broker workflows. These results in two possible types for the `info` parameter.

This PR updates the type definitions to be more specific about the types for the `info` parameter for the broker's `errorHandler` option.

### :dart: Relevant issues
None

### :gem: Type of change

I patched these changes into my repository and then created a new error handler function which used the `info` parameter.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -985,7 +985,11 @@ declare namespace Moleculer {
 		nodeId?: string;
 	}
 	type BrokerErrorHandlerInfo = BrokerErrorHandlerInfoAction | BrokerErrorHandlerInfoBroker;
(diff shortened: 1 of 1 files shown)


icebob (approved):
Great, thanks!

Issue #1296, opened 2024-07-28
Memory leak when a call to an action stream fails


Thalandor:
## :stop_sign: Do you want to ask a question ?

Please head to the [Discord chat](https://discord.gg/TSEcDRP)

## Prerequisites

Please answer the following questions for yourself before submitting an issue. 

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository

## Current Behavior

When a call to an action stream fails for whatever reason from the service listener, node memory increases non-stop. The more calls you make to the action stream and fails, the more the memory increases until you have an out of memory exception.

## Expected Behavior

When an action stream fails, memory does not increase on the receptor side.

## Failure Information

Checked on moleculer 14 and moleculer 15.

### Steps to Reproduce

1. Create two services, a caller and a listener
2. Listener implements an action stream
3. Caller calls the listener action stream with a big file (test done with a 50mb file)
4. Make listener fail (for example making the action stream just throw an exception) and memory is not released.
5. Caller calls again
6. Listener fails again
7. After several retries listener throws a out of memory exception.

### Reproduce code snippet
I created this repository with a minimal example and a readme to reproduce it
https://github.com/Thalandor/moleculer-issue

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Moleculer version: Tried with 14.33 and 15-beta1
* NodeJS version: 20
* Operating System: Ubuntu
* Services running on docker images. The repo does not have the docker images done, however the memory is increased anyway.

Thib3133 from the discord server provides some hints regarding the possible error:
![image](https://github.com/user-attachments/assets/07736f82-9001-4fbe-a822-6c2d490a8de5)

### Failure Logs

![image](https://github.com/user-attachments/assets/f9704524-089f-4c28-abe3-9b4d9461906a)

icebob:
Pls check it, in 0.14.35. If it still alive, please reopen.

PR #1306, opened 2024-11-05
fix: updated removePendingRequestByNodeID

JS-AK:
added clean and close this.pendingReqStreams when node disconnected

## :memo: Description
Fixed memory leak when another microservice started stream with broker.call and then switched off (with reboot for example)

Create two services, a caller and a listener
Listener implements an action stream
Caller calls the listener action stream
Make Caller fail (restart microservice) and Listener memory/stream (in src/transit.js -> this.pendingReqStreams) is not released.
Caller calls again
Caller fails again
After several retries listener may throws a out of memory exception.

PS 
this.pendingRequests was **empty** on Listener microservice 

### :gem: Type of change

Repo for reproduce this bug
https://github.com/JS-AK/moleculer-pr-1306-example

### DIFF
--- a/src/transit.js
+++ b/src/transit.js
@@ -548,7 +548,7 @@ class Transit {
 			pass.$prevSeq = -1;
 			pass.$pool = new Map();
 
-			this.pendingReqStreams.set(payload.id, pass);
+			this.pendingReqStreams.set(payload.id, { sender: payload.sender, stream: pass });
 		}
 
 		if (payload.seq > pass.$prevSeq + 1) {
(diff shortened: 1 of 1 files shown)


icebob (approved):
Good job, thanks!

icebob:
@JS-AK what do you think, it can resolve https://github.com/moleculerjs/moleculer/issues/1296 as well?

JS-AK:
> @JS-AK what do you think, it can resolve #1296 as well?

it looks like that

icebob:
Released.

PR #1318, opened 2025-01-27
Add new option for initial services dependencies

GabrielButarelo:
## :memo: Description

At the company where I work, we recently encountered a challenge with MoleculerJS. We needed to initialize a service responsible for managing database connections before all the other services we have built (which currently number in the hundreds). The solution we found was to add the dependencies field to the schema of every service, referencing the name of the service we wanted the others to wait for. This led to the development of an improvement where we can pass the --dependencies option in the runner, specifying the names of the services that all other services should wait for to load. Alternatively, we can simply add them to the initialDependencies array in the broker.

### :dart: Relevant issues
N/A

### :gem: Type of change

### :scroll: Example code
```js
    initialDependencies: ['databases-service']
});
``` 

ts-node ./node_modules/moleculer/bin/moleculer-runner.js --hot --repl --config moleculer.config.ts services/**/*.services.ts --dependencies databases-service

To test it, simply create two services and set one as a dependency. If you're using the runner, you can pass the --dependencies option. If you're manually instantiating the broker, you can use the initialDependencies option in the settings. Run the program and check if it waits for the dependent service to start before starting the other.

### DIFF
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -86,7 +86,8 @@ export default class MoleculerRunner {
 			.option("env", "Load .env file from the current directory")
 			.option("envfile", "Load a specified .env file")
 			.option("instances", "Launch [number] instances node (load balanced)")
-			.option("mask", "Filemask for service loading");
+			.option("mask", "Filemask for service loading")
+			.option("dependencies", "List of global dependencies for all services");
 
 		this.flags = Args.parse(procArgs, {
 			mri: {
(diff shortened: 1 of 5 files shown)


icebob:
What is the use-case which is not solved by service dependencies but solved by your initialDependencies?

GabrielButarelo:
Regular service dependencies require specifying dependencies in each service, meaning any update (such as adding or modifying a core dependency like a database service) would require changes across all service files.

initialDependencies allows defining common dependencies globally, reducing redundancy and making maintenance easier in large-scale projects.

This feature improves maintainability and consistency across many services without requiring individual modifications.

PetrChalov:
@GabrielButarelo This problem is solved more easily:

```
// moleculer.config.js
module.exports = {
  middlewares: [
    {
      serviceCreating(_service, schema) {
        schema.dependencies = ['databases-service', ...(schema.dependencies || [])];
      },
    },
  ],
};
```

PR #1342, opened 2025-12-12
Forward termination signals to cluster workers in runner

JS-AK:
Problem: when running with --instances in Docker/K8s, cluster workers didn’t receive SIGTERM, so they skipped graceful shutdown.
Fix: on stop signals, forward the signal to each cluster.worker before cluster.disconnect in runner.js and runner-esm.mjs.
Scope: stop handling only; no other behavior changes.

### DIFF
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -474,6 +474,11 @@ export default class MoleculerRunner {
 			process.on(signal, () => {
 				logger.info(`Got ${signal}, stopping workers...`);
 				stopping = true;
+				// Forward termination signals to workers so they can cleanup gracefully
+				for (const id in cluster.workers) {
+					logger.info(`Forwarding ${signal} to worker ${id}...`);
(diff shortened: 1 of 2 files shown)


icebob:
Thank you for the fix! 👍

I verified it: with `--instances`, when the stop signal is delivered only to the master process (Docker/K8s semantics), the workers' brokers indeed never stopped gracefully — they just kept running until they would be SIGKILL-ed. Your signal forwarding fixes this properly.

One platform note: on Windows `process.kill()` doesn't deliver a catchable signal — it terminates the target process immediately, without running its signal handlers. So there the forwarding would actually *abort* the workers' graceful shutdown (on Windows the workers receive the console Ctrl+C event directly anyway, so forwarding isn't needed). Since it's not necessary on Windows, I've pushed a small commit to the PR (2660dc9c) that skips the forwarding on `win32`.

Thanks again for the contribution!

PR #1344, opened 2025-12-19
Update mergeOptions signature

fstanis:
This fixes the type definition of `mergeOptions` to match what's in the code:

https://github.com/moleculerjs/moleculer/blob/master/src/runner.js#L315

> **Note:** originally this patched `index.d.ts`. Since then the typings were split out into separate `src/*.d.ts` files, so the change now applies to [`src/runner.d.ts`](https://github.com/moleculerjs/moleculer/blob/master/src/runner.d.ts#L117). Rebased onto `master` by a maintainer; the fix itself is unchanged.

### DIFF
--- a/src/runner.d.ts
+++ b/src/runner.d.ts
@@ -114,7 +114,7 @@ declare class Runner {
 	/**
 	 * Merge broker options from config file & env variables
 	 */
-	mergeOptions(): void;
+	mergeOptions(configFromFile: Partial<BrokerOptions>): void;
 
 	/**
 	 * Check if a path is a directory


icebob-ai:
Thanks for the fix, and sorry for the long delay! 🙏

The change is correct: `loadConfigFile()` always resolves an object (`{}` when no config file is found), and the only caller (`runner.js:543`) always passes it — so a non-optional `configFromFile: Partial<BrokerOptions>` is exactly right.

In the meantime the typings were split out of the monolithic `index.d.ts`, so the `Runner` declaration now lives in [`src/runner.d.ts`](https://github.com/moleculerjs/moleculer/blob/master/src/runner.d.ts#L117) — which is why this went into a conflicting state. I've rebased your branch onto `master` and moved the one-liner to the new location. The fix itself is unchanged and your authorship is kept on the commit. `BrokerOptions` was already imported there, so no extra changes were needed.

`npm run tsd` and `npm run test:ts` pass locally. Should be good to go once CI is green.

Thanks again for catching this!

PR #1355, opened 2026-05-07
fix: Allow to set false for actions without mixins

Embraser01:
## :memo: Description

This allows to easily have "conditional" actions defined in your service (cf. exemple).
The alternative is to wrap it with an object spread or have a "false" mixin handler.

### :dart: Relevant issues

I didn't find any issue about that, but I can create one if you prefer.

### :gem: Type of change

### :scroll: Example code

```ts
export default {
	name: 'test',
	actions: {
		sum: !env.ENABLE_SUM && {
			handler(ctx) {
				return ctx.params.a + ctx.params.b
			}
		}
	}
}
```

We got the issue on our side and found the `mergeSchemaActions` method that handles it. I then found a test case to
add this use case to see if I could repro it.
Once reproduced in unit tests, I fixed it in the code.

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -716,7 +716,7 @@ class Service {
 	 */
 	mergeSchemaActions(src, target) {
 		Object.keys(src).forEach(k => {
-			if (src[k] === false && target[k]) {
+			if (src[k] === false) {
 				delete target[k];
 				return;
 			}
--- a/test/unit/service.spec.js
+++ b/test/unit/service.spec.js
@@ -1781,7 +1781,8 @@ describe("Test Service class", () => {
 					},
 					handler() {}
 				},
-				update: false
+				update: false,
+				customFind: false
 			};
 
 			const prev = {


icebob-ai:
Thanks for catching this — the fix is correct and we'll merge it. 🙏

One clarification for the record, because the repro snippet in the description is slightly off and could confuse people who find this issue later.

The example service has **no mixins at all**, and in that case `actions: { sum: false }` already works on `master`. `mergeSchemaActions()` is only reachable through `applyMixins()` ([`service.js:78`](https://github.com/moleculerjs/moleculer/blob/master/src/service.js#L78)), which is skipped entirely when `schema.mixins` is missing — and it also early-returns on an empty array (`mixins.length > 0`).

The actual trigger is: the service **has** at least one mixin, and the `false`-valued action has **no counterpart in any of them**. Test matrix on `master`:

| schema | result |
| --- | --- |
| no `mixins` key | ✅ works |
| `mixins: []` | ✅ works |
| `mixins: [{}]` | 💥 `ServiceSchemaError` |
| `mixins: [{ actions: { other() {} } }]` | 💥 `ServiceSchemaError` |
| `mixins: [{ actions: { sum() {} } }]` | ✅ works (caught by the old `&& target[k]` guard) |

The root cause is worth writing down as well. Without your fix, the `false` falls through to `target[k] = _.defaultsDeep(srcAction, targetAction)`, and lodash returns a **boxed** `Boolean` object there rather than the primitive:

```js
typeof _.defaultsDeep(false, undefined); // "object" → [Boolean: false]
```

So the `if (action === false) return;` guard in `parseServiceSchema()` ([`service.js:169`](https://github.com/moleculerjs/moleculer/blob/master/src/service.js#L169)) no longer matches, `_createAction()` gets handed the boxed object, and you end up with `ServiceSchemaError: Missing action handler on 'sum' action in '...' service!`. Your change avoids that path altogether, and since `delete target[k]` on a missing key is a no-op, dropping the `&& target[k]` condition is safe.

For what it's worth, your reading of "without mixins" (i.e. *without a mixin definition for that action*) is consistent with the test you added — `customFind: false` has no counterpart in `prev`, whereas `update: false` does and already worked before. So only the example snippet needs fixing: it should include a `mixins: [...]` entry to actually reproduce.

Full unit suite passes locally. Merging — thanks again for the fix and for the test coverage!

PR #1362, opened 2026-07-07
fix(registry): merge actions/events on grouping and emit servicesChanged on disconnect

JS-AK:
## :memo: Description

Two related bugs that cause API Gateway `autoAliases` to return 404 for actions that exist on a live node when a stale node (zombie) is still present in the registry (e.g. a Kubernetes pod killed via SIGKILL without SIGTERM):

**1. `ServiceCatalog.list({ grouping: true })` drops actions/events from non-first nodes**

When multiple nodes run the same service, `list()` with `grouping: true` only kept actions/events from the **first** `ServiceItem` found. Actions from all subsequent nodes were silently discarded.
This means if a zombie pod (killed via SIGKILL, still in registry) was processed first, any new actions added by a replacement pod were invisible to API Gateway — resulting in permanent 404s until full restart.

**Fixed by:** merging the union of actions/events from all nodes into the grouped result (first-node wins on conflict, protected actions and internal `$`-events are excluded).

**2. `NodeCatalog.disconnected()` never calls `broker.servicesChanged()`**

When a node disconnects (heartbeat timeout), services were removed from the catalog and `$node.disconnected` was emitted — but `broker.servicesChanged()` was never called. This meant API Gateway never regenerated its `autoAliases` route table after a disconnect, leaving stale 404 routes indefinitely.

**Fixed by:** adding `broker.servicesChanged(false)` after the disconnect is processed in `NodeCatalog.disconnected()`.

### :dart: Relevant issues

Related to any environment where pods/nodes can die via SIGKILL (no SIGTERM) — common in Kubernetes.

### :gem: Type of change

### :scroll: Example code

```js
// Scenario: two nodes run the same service, zombie pod was registered first
// Before fix: API Gateway only saw actions from pod-OLD, svc.newAction → 404
// After fix: actions from all nodes are merged into the grouped result

const broker = new ServiceBroker({ nodeID: "gateway" });

broker.createService({
  name: "api",
  mixins: [ApiGateway],
  settings: {
    routes: [{ autoAliases: true }]
  }
});

// pod-OLD (zombie, still in registry after SIGKILL)
// pod-NEW (replacement, has extra action svc.newAction)
// After fix: GET /svc/new-action → 200 ✓
```

- [x] **Unit — `service-catalog.spec.js`**: regression test for zombie pod scenario (merge of new actions/events from second node); edge-case tests for protected actions not merged, internal `$`-events not merged, first-node action not overwritten
- [x] **Unit — `node-catalog.spec.js`**: verified `broker.servicesChanged(false)` is called on both expected and unexpected disconnect
- [x] **Integration — `registry.spec.js`**: updated assertion to expect 2 `servicesChanged` calls after node stop (INFO packet + DISCONNECT packet)

### DIFF
--- a/src/registry/node-catalog.js
+++ b/src/registry/node-catalog.js
@@ -203,6 +203,8 @@ class NodeCatalog {
 
 			this.broker.broadcastLocal("$node.disconnected", { node, unexpected: !!isUnexpected });
 
+			this.broker.servicesChanged(false);
+
 			this.registry.updateMetrics();
 
 			if (isUnexpected) this.logger.warn(`Node '${node.id}' disconnected unexpectedly.`);
--- a/src/registry/service-catalog.js
+++ b/src/registry/service-catalog.js
@@ -159,6 +159,24 @@ class ServiceCatalog {
 				res.push(item);
 			} else {
 				if (item.nodes.indexOf(service.node.id) === -1) item.nodes.push(service.node.id);
+
+				// Merge actions from subsequent nodes so that autoAliases always
+				// sees the full union of actions across all nodes for this service.
+				// This prevents stale registry entries (e.g. a k8s zombie pod killed
+				// via SIGKILL) from shadowing newer actions of a replacement pod.
+				if (withActions && item.actions) {
+					_.forIn(service.actions, (action) => {
+						if (action.protected || item.actions[action.name]) return;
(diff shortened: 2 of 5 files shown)


icebob:
Nice catch, thanks!

PR #1364, opened 2026-08-05
feat(runner): support worker node args and auto heap balance

JS-AK:
## :memo: Description

Adds ~~two~~ Moleculer Runner option~~s~~ for cluster mode (`--instances`):

1. **`--worker-node-args`** / `MOLECULER_WORKER_NODE_OPTIONS`  
   Pass arbitrary Node.js/V8 CLI flags only to forked workers via `cluster.setupPrimary({ execArgv })`. The primary process is not affected. This avoids putting worker-only flags into global `NODE_OPTIONS`.

2. ~~**`--auto-worker-heap-reserve=<MB>`** / `MOLECULER_AUTO_WORKER_HEAP_RESERVE`  
   Reserve N MB for the primary (supervisor) process, detect available memory (`process.constrainedMemory()` in containers / cgroup, otherwise `os.totalmem()`), and split the remainder equally across workers as `--max-old-space-size`.~~

~~**Precedence:** when `--auto-worker-heap-reserve` is set, it overrides any `--max-old-space-size*` / `--max-old-space-size-percentage*` values from `--worker-node-args`. Other worker args are kept.~~

**Notes for reviewers:**
- With `--instances`, the primary does **not** start a broker — it only forks/restarts workers.
- ~~`--max-old-space-size` is a V8 heap limit, not a hard RSS/container limit.~~
- Same changes are applied to both CJS (`runner.js`) and ESM (`runner-esm.mjs`) runners.

### :dart: Relevant issues

N/A (feature request / enhancement)

### :gem: Type of change

### :scroll: Example code

```bash
# Manual heap limit per worker
moleculer-runner -i 3 --worker-node-args="--max-old-space-size=768" services

# Via env
MOLECULER_WORKER_NODE_OPTIONS="--trace-warnings --max-old-space-size=768" moleculer-runner -i 3 services
```

~~Auto-balance (removed): reserve 1024 MB for primary, split the rest equally~~  
~~`moleculer-runner -i 3 --auto-worker-heap-reserve=1024 services`~~  
~~`MOLECULER_AUTO_WORKER_HEAP_RESERVE=1024 moleculer-runner -i 3 services`~~  
~~`MOLECULER_WORKER_NODE_OPTIONS="--trace-warnings" moleculer-runner -i 3 --auto-worker-heap-reserve=1024 services`~~

- [x] **Unit — `test/unit/runner.spec.js`**: parse `--worker-node-args` / `-w`; merge env + CLI; `cluster.setupPrimary({ execArgv })` called for workers ~~; parse `--auto-worker-heap-reserve`; equal heap split; override of conflicting heap flags; invalid reserve error~~
- [x] Locally: `npx jest test/unit/runner.spec.js --coverage=false` (~~17~~ 11 tests passed)

### DIFF
--- a/src/runner-esm.mjs
+++ b/src/runner-esm.mjs
@@ -78,7 +78,6 @@ export default class MoleculerRunner {
 		-s, --silent     Silent mode. No logger (disabled by default)
 		-v, --version    Output the version number
 		-w, --worker-node-args  Node.js/V8 args for cluster workers only
-		    --auto-worker-heap-reserve  MB for primary; auto-split remaining available RAM equally across workers
 	*/
 	processFlags(procArgs) {
 		Args.option("config", "Load the configuration from a file")
@@ -92,10 +91,6 @@ export default class MoleculerRunner {
 			.option(
 				"worker-node-args",
 				"Node.js/V8 args passed to cluster workers only (also MOLECULER_WORKER_NODE_OPTIONS)"
-			)
-			.option(
-				"auto-worker-heap-reserve",
-				"Auto-balance worker heaps: MB reserved for primary; remaining available RAM is split equally as --max-old-space-size (also MOLECULER_AUTO_WORKER_HEAP_RESERVE)"
 			);
 
 		this.flags = Args.parse(procArgs, {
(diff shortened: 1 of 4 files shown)


JS-AK:
## Coveralls note (−2.4%)

Coverage dropped because unit tests now `require` `src/runner.js`, so Coveralls includes the **entire** runner file in the report.

Previously this file was almost never exercised by the unit suite, so it did not meaningfully affect the aggregate %. Pulling it in at ~25% file coverage lowers the overall number even though the **new** logic is largely covered (Coveralls: **39/43 new lines**, 90.7%).

The only uncovered *new* lines on that build were:
- the **validation error branch** for `--auto-worker-heap-reserve` (when the value is not a number or is negative -> `throw`)
- `getAvailableMemoryMb()` body (tests mocked this method, so the real implementation was not executed)

src/runner-esm.mjs
```diff
@@ -98,16 +108,155 @@ export default class MoleculerRunner {
...
 	}
 
+	/**
+	 * Parse a Node.js CLI args string into an argv array.
+	 * Supports quoted tokens, same style as NODE_OPTIONS.
+	 *
+	 * @param {string|string[]|null|undefined} value
+	 * @returns {string[]}
+	 */
+	parseNodeArgsString(value) {
```
icebob, line 124: Please check to parse the new parameter with the existing `Args.parse(...)`, because it would be easier than implement a parsing solution again.

JS-AK: Args.parse already handles --worker-node-args as one string value. What’s left is tokenizing that string (and the env var) into an execArgv array. Args.parse can’t do that - it only knows registered runner options and returns a flags object, not Node argv tokens. Unknown Node flags (e.g. --max-old-space-size) are rejected as unknown options.

JS-AK:
After Discord discussion, auto-worker-heap-reserve was removed; only --worker-node-args is kept.

